PR #3, opened 2021-07-28
HEC-779-782-286: auth & IV

AndyHWChung:


app/uk/gov/hmrc/hecapplicantfrontend/util/HttpResponseOps.scala
```diff
@@ -0,0 +1,54 @@
...
+    def parseJSON[A](
+      path: Option[String] = None
+    )(implicit reads: Reads[A]): Either[String, A] =
+      Try(
+        path.fold[JsLookupResult](JsDefined(response.json))(response.json \ _)
+      ) match {
+        case Success(jsLookupResult) ⇒
+          // use Option here to filter out null values
+          jsLookupResult.toOption
+            .flatMap(Option(_))
```
priyankadhiman3006, line 37: I think, .flatMap(Option(_)) is not needed here, it's just creating the same values as jsLookupResult.toOption i.e. Option[JsValue]

AndyHWChung: it was a while ago when I wrote this (I borrowed it from a previous project), but I think I put the `flatMap` in cos `jsLookupResult` was returning `Some(null)` which we actually want to treat as a `None`. It would only happen if `path` in `parseJSON` was defined, but I don't think we're explicitly passing in that variable anywhere so we could probably even just remove that input parameter

AndyHWChung: updated

AndyHWChung:
> can we do something like this here:
> .fold[Either[String, A]](
> response ⇒
> // there was JSON in the response but we couldn't read it
> Left(s"Could not parse http response JSON $response"),
> Right(response)
> )
> Sometime there is error in json, where a field is giving parsing error and we need to know what exactly is wrong with json. In that scenario, it's good to get the error message which can tell which field is causing the json to fail the parse method. If this scenario doesn't matter here, please feel free to ignore this comment

it would be easier to have it there but when we go to production we don't want to accidentally log personal/sensitive data for everyone to see in kibana. If it turns out that we can't parse the data we would have to go to a secure area to inspect the payload to see what went wrong. You'd need security clearance to be able to see that data in prod.

test/uk/gov/hmrc/hecapplicantfrontend/controllers/AuthAndSessionDataBehaviour.scala
```diff
@@ -0,0 +1,111 @@
...
+
+  def authBehaviour(performAction: () => Future[Result]): Unit = {
+    "redirect to the login page when the user is not logged in" in {
+      List[NoActiveSession](
+        BearerTokenExpired(),
+        MissingBearerToken(),
+        InvalidBearerToken(),
+        SessionRecordNotFound()
+      ).foreach { e =>
+        withClue(s"For AuhtorisationException $e: ") {
```
priyankadhiman3006, line 59: Typo in AuhtorisationException

test/uk/gov/hmrc/hecapplicantfrontend/controllers/ControllerSpec.scala
```diff
@@ -0,0 +1,142 @@
...
+          ConfigFactory.parseString(
+            """
+            | microservice.metrics.graphite.enabled = false 
+      | """.stripMargin
+          )
+        ).withFallback(additionalConfig)
+      )
+      .disable[uk.gov.hmrc.hecapplicantfrontend.repos.SessionStore]
+      .overrides(overrideBindings: _*)
+      .overrides(bind[MessagesApi].toProvider[TestMessagesApiProvider])
```
priyankadhiman3006, line 59: Out of curiosity, why we are not adding this binding to the def overrideBindings above at line 42?

priyankadhiman3006 (approved):
I have added a few comments /questions. Feel free to answer them when you have time. Rest looks good. As of now, approving it.

app/uk/gov/hmrc/hecapplicantfrontend/config/AppConfig.scala
```diff
@@ -16,13 +16,61 @@
...
+  lazy val redirectToIvUplift: Result = {
+    val ivUrl: String = platformHost.getOrElse(config.get[String]("iv.url"))
+
+    val ivOrigin: String = config.get[String]("iv.origin")
+
+    val (ivSuccessUrl: String, ivFailureUrl: String) = {
+      val useRelativeUrls                          = platformHost.isDefined
+      val (successRelativeUrl, failureRelativeUrl) =
+        routes.StartController.start().url ->
+          routes.IvFailureController.ivFailure(UUID.randomUUID()).url.takeWhile(_ =!= '?')
```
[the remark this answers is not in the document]
AndyHWChung: in general I try to avoid doing unsafe getting by an index in case it throws an exception - with your example it won't, but it still scares me a bit to see getting by an index :D

app/uk/gov/hmrc/hecapplicantfrontend/controllers/actions/SessionDataAction.scala
```diff
@@ -0,0 +1,65 @@
...
+  override def refine[A](
+    request: AuthenticatedRequest[A]
+  ): Future[Either[Result, RequestWithSessionData[A]]] = {
+    implicit val hc: HeaderCarrier =
+      HeaderCarrierConverter.fromRequestAndSession(request, request.session)
+
+    sessionStore
+      .get()
+      .leftMap { e =>
+        logger.warn("Could not get session data", e)
```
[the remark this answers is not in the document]
AndyHWChung: as long as we're careful when creating `String`  version of `Error` that side should be safe - but we might accidentally slip when doing `.getMessage` on the `Throwable` on that side of the `Error` so it might be safer to just print out the name of the exception. However, then we lose potentially valuable info on the throwable cos `"Could not get session data: RuntimeException"` in not particularly helpful. In practice, the only place we're creating the `Thowable` versions of `Error` are when we catch exceptions in the connectors, and the exceptions from there shouldn't be logging sensitive info

app/uk/gov/hmrc/hecapplicantfrontend/services/CitizenDetailsService.scala
```diff
@@ -0,0 +1,111 @@
...
+        httpResponse
+          .parseJSON[CidPerson]
+          .leftMap(Error(_))
+          .flatMap(toCitizenDetails)
+
+    }
+
+  @SuppressWarnings(Array("org.wartremover.warts.Any"))
+  private def toCitizenDetails(cidPerson: CidPerson): Either[Error, CitizenDetails] = {
+    val nameValidation: ValidatedNel[String, Name] =
```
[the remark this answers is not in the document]
AndyHWChung: at the moment I think this is probably the only place we'll use this kind of validation - we're not going to be reading names and dob's from any other sources. If it turns out we do we can always move this stuff out when it happens

app/uk/gov/hmrc/hecapplicantfrontend/util/HttpResponseOps.scala
```diff
@@ -0,0 +1,47 @@
...
+import uk.gov.hmrc.http.HttpResponse
+
+import scala.util.{Failure, Success, Try}
+
+object HttpResponseOps {
+
+  implicit class HttpResponseOps(private val response: HttpResponse) extends AnyVal {
+
+    def parseJSON[A](implicit reads: Reads[A]): Either[String, A] =
+      // use Option here to filter out null values
```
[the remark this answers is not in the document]
AndyHWChung: good point, there is no option now

test/uk/gov/hmrc/hecapplicantfrontend/controllers/ControllerSpec.scala
AndyHWChung: no particular reason other than it'll look slightly uglier cos you'd have to do something like:
```scala
.overrides((overrideBindings :+ bind[MessagesApi].toProvider[TestMessagesApiProvider]): _*)
```

AndyHWChung: oh, just read what you wrote properly - it's to basically let classes implementing `ControllerSpec` to not worry about this specific binding. Otherwise, each implementing class would have to worry about using `super.overrideBindings` which is kind of annoying

PR #9, opened 2021-08-11
HEC- 784: license expiry date

AndyHWChung:


test/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswersSpec.scala
```diff
@@ -59,8 +66,11 @@ class UserAnswersSpec extends AnyWordSpec with Matchers {
 
       "unsets the licence type field" in {
```
priyankadhiman3006, line 67: Should we say unsets the licence type and expiry date field?

priyankadhiman3006 (approved):

test/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswersSpec.scala
AndyHWChung: oh yeah - my intention here was to have separate parts of the specs to cover the different fields, so expiry date unstting would be done in a separate `in` block

AndyHWChung: that way we have tests which are generally testing one thing at a time

AndyHWChung:
LGTM

PR #13, opened 2021-08-17
Hec 788 applicant tax situation

priyankadhiman3006:


app/uk/gov/hmrc/hecapplicantfrontend/models/TaxDisplayYear.scala
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package uk.gov.hmrc.hecapplicantfrontend.models
+
+import play.api.libs.json.Format
+import play.api.libs.functional.syntax.toInvariantFunctorOps
+
+final case class TaxDisplayYear(year: Int) extends AnyVal
```
AndyHWChung, line 22: a more general name like `TaxYear` would open this case class up to use for other purposes if we need it in the future for something other than displaying

app/uk/gov/hmrc/hecapplicantfrontend/util/TimeUtils.scala
```diff
@@ -125,4 +126,15 @@ object TimeUtils {
 
   }
 
+  def getTaxYearDisplayDate(d: LocalDate): TaxDisplayYear = {
```
AndyHWChung, line 129: this might be better off closer to where it's being called - the business logic here is quite specific and it's not clear what `display` refers to here in isolation.

Something like 
```scala
def getTaxYear(date: LocalDate): TaxYear = ???
```
is quite general and would belong in here though and could help you with the implementation of the business logic. 

`currentYearTaxStartDate` may not return what you expect if  `Jan 1st <= today < 6th April`

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -16,23 +16,118 @@
...
-  mcc: MessagesControllerComponents
-) extends FrontendController(mcc) {
+  mcc: MessagesControllerComponents,
+  taxSituationPage: html.TaxSituation
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc)
+    with I18nSupport
+    with Logging {
+
+  val currentdate: LocalDate = TimeUtils.today()
```
AndyHWChung, line 51: `TimeUtils.today` should be called as late as possible where it's needed - as a `val` here, it'd evaluate a date and it'd be wrong the next day if the service does not stop in between

app/uk/gov/hmrc/hecapplicantfrontend/util/TimeUtils.scala
priyankadhiman3006: Agree with the method name , it can be more general. But suppose if today is 1 jan 2021, 6 months back is 1 july 2020, so the tax year which ended  before 1 july 2020 is 2019 to 2020, so the code will give 2019 as value. Please let me know if i understand it wrong?

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
priyankadhiman3006: yes, good point 👍

app/uk/gov/hmrc/hecapplicantfrontend/models/TaxYear.scala
```diff
@@ -0,0 +1,29 @@
...
+package uk.gov.hmrc.hecapplicantfrontend.models
+
+import play.api.libs.json.Format
+import play.api.libs.functional.syntax.toInvariantFunctorOps
+
+final case class TaxYear(year: Int) extends AnyVal
+
+object TaxYear {
+
+  implicit val format: Format[TaxYear] =
```
AndyHWChung, line 26: we don't need to store the tax year anywhere so we could get away with not having a format instance for now

test/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationControllerSpec.scala
```diff
@@ -0,0 +1,583 @@
...
+
+  override def overrideBindings = List(
+    bind[AuthConnector].toInstance(mockAuthConnector),
+    bind[SessionStore].toInstance(mockSessionStore),
+    bind[JourneyService].toInstance(mockJourneyService),
+    bind[TimeProvider].toInstance(mockTimeProvider)
+  )
+
+  def mockTimeProviderToday(d: LocalDate)          = (mockTimeProvider.currentDate _).expects().returning(d)
+  def taxYearMessage(startYear: Int, endYear: Int) = s"This is for the tax year $startYear to $endYear."
```
AndyHWChung, line 54: to future proof this, this could be:
```scala
messageFromMessageKey("taxSituation.hint", startYear.toString, endYear.toString)
```
If the message ever changes then we don't need to update this spec this way

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -16,23 +16,125 @@
...
+  def taxSituationForm(options: List[TaxSituation]): Form[TaxSituation] =
+    Form(
+      mapping(
+        "taxSituation" -> of(FormUtils.radioFormFormatter(options))
+      )(identity)(Some(_))
     )
+
+  def getTaxYear(currentDate: LocalDate): TaxYear = {
+    val currentYear             = currentDate.getYear
+    val currentYearTaxStartDate = LocalDate.of(currentYear, 4, 6)
```
AndyHWChung, line 134: if today was 1st Jan 2021, this logic would have:
```scala
val currentYear = 2021
val currentYearTaxStartDate = 6th April 2021
```
but the start date of the tax year that 1st Jan 2021 falls in is 6th April 2020. Would that affect the logic you have here?
Isn't the business rule more like this?
```scala
  def taxYear(date: LocalDate): TaxYear = {
    val currentCalendarTaxYearStart = LocalDate.of(date.getYear, 4, 6)
    if (date < currentCalendarTaxYearStart) TaxYear(date.getYear - 1)
    else TaxYear(date.getYear)
  }

 def getDisplayedTaxYear(): TaxYear = {
  val today = timeProvider.today()
  val taxYearOfSixMonthsAgo = taxYear(today.minusMonths(6L))
  TaxYear(taxYearOfSixMonthsAgo.year - 1)
} 
```

priyankadhiman3006: I think we are both doing the same logic with different way, in my logic i am doing 
 if (sixMonthEarlierDate.isBefore(currentYearTaxStartDate)) TaxYear(currentYear - 2)
    else TaxYear(currentYear - 1)
    in yours your are finding the year of the date which is six month earlier , then subtracting 1. 
    We can do either way , result is same. :)
    i am happy to do it your way as well

priyankadhiman3006: Besides, I think in your logic it should be  if (date > currentCalendarTaxYearStart) TaxYear(date.getYear - 1)

AndyHWChung: time is so confusing! If the test cases pass then that does prove that your logic is correct - in my head it's just difficult to equate the two for some reason :D

priyankadhiman3006: yeah, we should have a call about this :)

priyankadhiman3006: i ran your logic with 13 aug 2021, it should give 2019, it gave TaxYear(2020)(it will be ok, if you are considering the end year, but for start year it should give 2019)

AndyHWChung:
LGTM

PR #14, opened 2021-08-19
HEC-789: CYA

AndyHWChung:


app/uk/gov/hmrc/hecapplicantfrontend/models/TaxDetails.scala
```diff
@@ -0,0 +1,39 @@
...
+import play.api.libs.json.{Json, OFormat}
+import uk.gov.hmrc.hecapplicantfrontend.models.ids.{CTUTR, NINO, SAUTR}
+
+sealed trait TaxDetails extends Product with Serializable
+
+object TaxDetails {
+
+  final case class IndividualTaxDetails(
+    nino: NINO,
+    sautr: Option[SAUTR]
```
priyankadhiman3006: i think tax situation is missing from here as well?

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -118,21 +128,71 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
         .getOrElse(sys.error(s"Could not find previous for $current"))
-    }
   }
 
-  private def licenceValidityPeriodRoute(session: HECSession): Call =
-    session.userAnswers.fold(_.licenceType, c => Some(c.licenceType)) match {
-      case Some(LicenceType.DriverOfTaxisAndPrivateHires) => routes.TaxSituationController.taxSituation()
-      case _                                              => routes.EntityTypeController.entityType()
-    }
+  private def licenceTypeForIndividualAndCompany(licenceType: LicenceType): Boolean = licenceType match {
```
priyankadhiman3006, line 144: can we reverse it and do something like this 
def isLicenceTypeForIndividualOnly(licenceType: LicenceType): Boolean = licenceType match {
case LicenceType.DriverOfTaxisAndPrivateHires => true
    case _                                        => false
  }
  In this way it make more sense to me, what do yo think? however, it's just my opinion, it can stay like this as well.

priyankadhiman3006 (changes requested):

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
AndyHWChung: in the future we may get licence type which are for companies only - with the underlying business rule being about licence types being applicable to both it may be safer to leave it like this

app/uk/gov/hmrc/hecapplicantfrontend/models/TaxDetails.scala
AndyHWChung: yeah, I updated it when you mentioned it on the other PR. Should be there now

nubz (approved):
Template refactoring looks good to me

priyankadhiman3006 (approved):

PR #19, opened 2021-08-31
HEC-789: don't wipe answers if licence type is same

AndyHWChung:


priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/controllers/LicenceDetailsController.scala
```diff
@@ -69,7 +69,10 @@ class LicenceDetailsController @Inject() (
   val licenceTypeSubmit: Action[AnyContent] = authAction.andThen(sessionDataAction).async { implicit request =>
     def handleValidLicenceType(licenceType: LicenceType): Future[Result] = {
       val updatedAnswers =
-        UserAnswers.empty.copy(licenceType = Some(licenceType))
+        if (request.sessionData.userAnswers.fold(_.licenceType, c => Some(c.licenceType)).contains(licenceType))
```
meghana-viswanath, line 72: was there a reason not to compare directly `c.licenceType == licenceType`?

meghana-viswanath (approved):

app/uk/gov/hmrc/hecapplicantfrontend/controllers/LicenceDetailsController.scala
AndyHWChung: if we did that check on the right then we'd have to do `_.licenceType.contains(licenceType)` on the left side of the fold which makes it a bit longer to read

PR #20, opened 2021-09-01
HEC-963: IF API integration

AndyHWChung:


AndyHWChung:
I stuck in validation for SAUTR's and CTUTR's using the domain library in this PR - if we're doing it in the BE we should catch invalid ones as early as possible to avoid users going through the journey and it breaking later on

meghana-viswanath (approved):

meghana-viswanath:
Do we need to add routes for these into `app.routes`?

AndyHWChung:
> Do we need to add routes for these into `app.routes`?

I didn't stick in any controller methods to expose any endpoints here - the code will get integrated with an actually endpoint when we come to implement the SA and CT pages in the journey

test/uk/gov/hmrc/hecapplicantfrontend/models/ids/SAUTRSpec.scala
```diff
@@ -32,6 +32,14 @@ class SAUTRSpec extends AnyWordSpec with Matchers {
       Json.fromJson[SAUTR](json) shouldBe JsSuccess(sautr)
     }
 
+    "have a method which validates an SA UTR from a string" in {
+      SAUTR.fromString("")            shouldBe None
```
priyankadhiman3006, line 36: can we include a test case for a 
1)blank space 
2)10 chars with alphanumeric values
pls ignore if not needed

AndyHWChung: more tests can't be a bad thing :D I made the validation a bit more lax with the spaces - I don't think we want to block journeys if we happen to get some id's with spaces in them. If we did, I'm guessing we'd take the action to just remove blank spaces if a user gets blocked anyways.

priyankadhiman3006 (approved):

PR #23, opened 2021-09-06
Display only relevant tax situation options based on licence type

meghana-viswanath:
Display only the relevant tax situation options for the `/tax-situation` route in `TaxSituation.scala.html` based on the licence type selected.

For `DriverOfTaxisAndPrivateHires`, display all options.
For all other licence types, display only `SA` and `NotChargeable` options.

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
@@ -22,7 +22,7 @@ import play.api.data.Form
 import play.api.data.Forms.{mapping, of}
 import play.api.i18n.I18nSupport
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents, Result}
-import uk.gov.hmrc.hecapplicantfrontend.controllers.TaxSituationController.{getTaxYear, taxSituationForm, taxSituationList}
+import uk.gov.hmrc.hecapplicantfrontend.controllers.TaxSituationController.{getTaxYear, taxSituationForm, taxSituationOptions}
 import uk.gov.hmrc.hecapplicantfrontend.controllers.actions.{AuthAction, SessionDataAction}
 import uk.gov.hmrc.hecapplicantfrontend.models.{TaxSituation, TaxYear}
 import uk.gov.hmrc.hecapplicantfrontend.models.TaxSituation._
@@ -31,8 +31,10 @@ import uk.gov.hmrc.hecapplicantfrontend.util.Logging.LoggerOps
 import uk.gov.hmrc.hecapplicantfrontend.util.{FormUtils, Logging, TimeProvider, TimeUtils}
 import uk.gov.hmrc.hecapplicantfrontend.views.html
 import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendController
-
 import java.time.LocalDate
+
+import uk.gov.hmrc.hecapplicantfrontend.models.licence.LicenceType
+
 import scala.concurrent.{ExecutionContext, Future}
 
 @Singleton
@@ -51,18 +53,17 @@ class TaxSituationController @Inject() (
   val taxSituation: Action[AnyContent] = authAction.andThen(sessionDataAction) { implicit request =>
     val licenceTypeOpt = request.sessionData.userAnswers.fold(_.licenceType, c => Some(c.licenceType))
     licenceTypeOpt match {
-      case Some(_) =>
+      case Some(licenceType) =>
         val back         = journeyService.previous(routes.TaxSituationController.taxSituation())
-        val reportIncome =
-          request.sessionData.userAnswers.fold(_.taxSituation, c => Some(c.taxSituation))
-        val options      = taxSituationList
+        val reportIncome = request.sessionData.userAnswers.fold(_.taxSituation, c => Some(c.taxSituation))
+        val options      = taxSituationOptions(licenceType)
         val form = {
           val emptyForm = taxSituationForm(options)
           reportIncome.fold(emptyForm)(emptyForm.fill)
         }
 
         Ok(taxSituationPage(form, back, options, getTaxYear(timeProvider.currentDate)))
-      case None    =>
+      case None              =>
         logger.error("Couldn't find licence Type")
         InternalServerError
     }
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -115,13 +116,21 @@ class TaxSituationController @Inject() (
...
     SAPAYE,
     NotChargeable
   )
 
+  def taxSituationOptions(licenceType: LicenceType): List[TaxSituation] =
+    licenceType match {
+      case LicenceType.DriverOfTaxisAndPrivateHires => allTaxSituations
+      case LicenceType.OperatorOfPrivateHireVehicles | LicenceType.ScrapMetalMobileCollector |
+          LicenceType.ScrapMetalDealerSite =>
+        List(SA, NotChargeable)
```
AndyHWChung, line 131: in order to not keep creating this list fresh multiple times we could extract it out in a `val` as well, something like:
```scala
private val allTaxSituations: List[TaxSituation] = List(
    PAYE,
    SA,
    SAPAYE,
    NotChargeable
  )

private val nonPAYETaxSituations = List(SA, NotChargeable)
```

test/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationControllerSpec.scala
```diff
@@ -420,6 +422,95 @@ class TaxSituationControllerSpec
...
+                mockTimeProviderToday(TimeUtils.today())
+              }
+
+              checkPageIsDisplayed(
+                performAction(),
+                messageFromMessageKey("taxSituation.title"),
+                { doc =>
+                  doc.select("#back").attr("href") shouldBe mockPreviousCall.url
+
+                  val options: Elements = doc.select(".govuk-radios__item")
```
AndyHWChung, line 494: we could potentially generalise this into a function to be made available in the `ControllerSpec` trait like:
```scala
def testRadioButtonText(doc: Document ,expectedRadioOptionsText: List[String]) = {
  val radioOptions = doc.select(".govuk-radios__item")
  radioOptions.size shouldBe expectedRadioOptionsText.size
  radioOptions.map(_.text)) shouldBe expectedRadioOptionsText
}
```
I think all of our radio's will probably have the class `.govuk-radios__item`. Then you could assert the 4 options in the text above as well.

AndyHWChung: looks like we're just asserting the size in other specs, but we could at some point enhance them to assert the text like you've done here as well

meghana-viswanath: I've just double-checked and there isn't a `map` operation available on `Elements`. But we can still do something like this maybe
```expectedRadioOptionsTexts.zipWithIndex.map({ case (text, i) => radioOptions.get(i).text() shouldBe text})```

Comment (meghana-viswanath) [reply test/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationControllerSpec.scala:494]: > looks like we're just asserting the size in other specs, but we could at some point enhance them to assert the text like you've done here as well

I can do as part of this PR given it's quite small?

Comment (AndyHWChung) [reply test/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationControllerSpec.scala:494]: if it doesn't take too long then you could go for it - don't want to increase the scope of your work too much! One option would be to extract out the method and use it for the page you're working on and try to get that merged in. Then someone else (like me) could go into the other specs for the other pages to use that method while you carry on with the rest of the ticket.

Comment (meghana-viswanath) [reply test/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationControllerSpec.scala]: 👍 yeah, that's what I have done, just limited it to the page I'm working on. (sorry, rebased instead of adding a new commit with changes)

**Review by AndyHWChung — APPROVED** [review]

File: app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala [hunk 3]
```diff
@@ -115,13 +116,23 @@ class TaxSituationController @Inject() (
...
     SAPAYE,
     NotChargeable
   )
 
+  private val nonPAYETaxSituations = List(SA, NotChargeable)
+
+  def taxSituationOptions(licenceType: LicenceType): List[TaxSituation] =
+    licenceType match {
+      case LicenceType.DriverOfTaxisAndPrivateHires => allTaxSituations
+      case LicenceType.OperatorOfPrivateHireVehicles | LicenceType.ScrapMetalMobileCollector |
```
priyankadhiman3006, line 131: instead of this , we can do case _ =>?

meghana-viswanath: You are right that this is more verbose than the catch all, but my intention was to avoid the catch all so that when we add a new licence type, it doesn't automatically match the catch all and throws a compile error. So we'll then consciously decide what the result should be for the new licence type.

priyankadhiman3006: I was thinking , in future even if we add new licence type, you will have to update this func, 
If that new licence type follow some  different behaviour, then you can add new case and the existing case_ will remain the same.
if the new license type follow same behaviour as other , then no need to update , case _ covered it all. What do you think?

meghana-viswanath: That's right, but there is a chance we might forget to update this function. So by being explicit, we make sure that the compiler throws an error and so acts as a reminder to update it. We can be sure we won't inadvertantly

priyankadhiman3006: Even if you forget , then your code will not work as expected and remind you . :) any how  it's ok if it is like this as well.

meghana-viswanath: Yeah, you are probably right. We'll have unit tests and automated tests to ensure this. But just on that off chance that we miss adding a unit test for that specific case, this will act as a reminder. I guess i just don't trust myself enough 😄

priyankadhiman3006 (approved):

PR #24, opened 2021-09-08
HEC-983 Tax situation page enhancements

meghana-viswanath:
Details in ticket - https://jira.tools.tax.service.gov.uk/browse/HEC-983

Note: URLs need confirmation.

Flow diagram to show journeys (dashed box shows scope of this ticket)

Screenshot 2021-09-08 at 11 10 46


app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -70,23 +71,52 @@ class TaxSituationController @Inject() (
...
-          request.sessionData.copy(userAnswers = updatedAnswers)
-        )
-        .fold(
-          { e =>
-            logger.warn("Could not update session and proceed", e)
-            InternalServerError
-          },
-          Redirect
-        )
+    val taxYear = getTaxYear(TimeUtils.today())
```
AndyHWChung, line 74: the tax year should be the one displayed on the `taxSituation` page which uses `getTaxYear` in the background - we should probably use `getTaxyear` here as well. There is an option to make the form submit the tax year in a hidden field but that opens up the service to users editing the html and modifying the tax year we use

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -70,23 +71,52 @@ class TaxSituationController @Inject() (
...
+              }
+            }) foldF ({ e =>
+              logger.warn("Failed to fetch SA status", e)
+              Future.successful(InternalServerError)
+            },
+            updateAndRedirect)
+          } else {
+            updateAndRedirect(request.sessionData.retrievedUserData)
+          }
+        case None      => Future.successful(Redirect(routes.TaxSituationController.sautrNotFoundExit()))
```
AndyHWChung, line 118: you may be able to get away with not having `Future.successful` here and above with the implicit conversion defined in the package object

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -70,23 +71,52 @@ class TaxSituationController @Inject() (
...
-        .fold(
-          { e =>
-            logger.warn("Could not update session and proceed", e)
-            InternalServerError
-          },
-          Redirect
-        )
+    val taxYear = getTaxYear(TimeUtils.today())
+
+    def handleValidTaxSituation(taxSituation: TaxSituation): Future[Result] = {
```
priyankadhiman3006, line 76: There are funcs inside funcs, it looks so confusing, will it be possible for you to separate them , so it's more clear to read?

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -70,23 +71,52 @@ class TaxSituationController @Inject() (
...
+      sautrOpt match {
+        case Some(utr) =>
+          if (Seq(TaxSituation.SA, TaxSituation.SAPAYE).contains(taxSituation)) {
+            (taxCheckService.getSAStatus(utr, taxYear) map { status =>
+              request.sessionData.retrievedUserData match {
+                case i: RetrievedApplicantData.IndividualRetrievedData => i.copy(saStatus = Some(status))
+                case c: RetrievedApplicantData.CompanyRetrievedData    => c
+              }
+            }) foldF ({ e =>
+              logger.warn("Failed to fetch SA status", e)
```
AndyHWChung, line 111: you could neaten this up with a for-comprehension to stack the `EitherT`s and avoid using `foldF` with something like:
```scala
val updatedAnswers =   request.sessionData.userAnswers.unset(_.taxSituation).copy(taxSituation = Some(taxSituation))

request.sessionData.retrievedUserData match {
  case individualRetrievedData: IndividualRetrievedData => // do this pattern match once
     val result = for{
       maybeSaStatus <-  if(saTaxSituations).contains(taxSituation)) // saTaxSituations in static list
                           individualRetrievedData.sautr
                              .map(taxCheckService.getSAStatus(_, taxYear))
                              .sequence  // push the option into the EitherT, probably need import  cats.instances.option._ for this
                         else EitherT.pure(None) 
       updatedRetrivedData = individualRetrievedData.copy(saStatus = maybeStatus)
       updatedAnswers =   request.sessionData.userAnswers.unset(_.taxSituation).copy(taxSituation = Some(taxSituation))
       next <- journeyService
                .updateAndNext(
                   routes.TaxSituationController.taxSituation(),
                   request.sessionData.copy(userAnswers = updatedAnswers, retrievedUserData = updatedRetrivedData)
                 )
       } yield next

     result.fold( e => ..., Redirect)


case _ => ... // error --- companies shouldn't see this page

   }

 i.sautr 
}
```
This way you only do the SA API call when you need to and the `JourneyService` maintains it's job as the sole thing that decides where the next page in the journey is

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -70,23 +71,52 @@ class TaxSituationController @Inject() (
...
+            Redirect
+          )
+      }
+
+      val sautrOpt = request.sessionData.retrievedUserData match {
+        case i: RetrievedApplicantData.IndividualRetrievedData => i.sautr
+        case _                                                 => None
+      }
+
+      sautrOpt match {
```
priyankadhiman3006, line 102: I think , if we put this in a separate func, then call that func at line 98, then you can get rid of the same logic again at line 107?

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -102,17 +132,38 @@ class TaxSituationController @Inject() (
                   formWithErrors,
                   journeyService.previous(routes.TaxSituationController.taxSituation()),
                   options,
-                  getTaxYear(TimeUtils.today())
+                  taxYear
```
priyankadhiman3006, line 135: I didn't store TimeUtils.today() in separate function as Andy mentioned in my PR that it could be risky if the year is changing by the time you all calculations, so it should be called as close as possible where needed.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -230,4 +230,23 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
     else routes.EntityTypeController.wrongGGAccount()
   }
 
+  private def taxSituationRoute(session: HECSession): Call =
+    (session.userAnswers match {
```
priyankadhiman3006, line 234: line 234- 237 can be put in a val and then map over that option, it will be more cleaner to read. what do you think?

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TechnicalExceptionController.scala
```diff
@@ -0,0 +1,39 @@
...
+
+@Singleton
+class TechnicalExceptionController @Inject() (
+  authAction: AuthAction,
+  sessionDataAction: SessionDataAction,
+  mcc: MessagesControllerComponents
+) extends FrontendController(mcc)
+    with Logging {
+
+  val technicalException: Action[AnyContent] = authAction.andThen(sessionDataAction) { implicit request =>
```
AndyHWChung, line 33: We don't need this as a separate page that we maintain - if we return `InternalServerError` then a standard platform page is returned via the `ErrorHandler` in the `config` package

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -230,4 +230,23 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
+      case i: IncompleteUserAnswers => i.taxSituation
+      case c: CompleteUserAnswers   => Some(c.taxSituation)
+    }) map ({
+      case TaxSituation.PAYE | TaxSituation.NotChargeable => routes.CheckYourAnswersController.checkYourAnswers()
+      case TaxSituation.SA | TaxSituation.SAPAYE          =>
+        session.retrievedUserData match {
```
AndyHWChung, line 240: there should be something here as well which maps onto the no sautr exit page if there is no sautr in session

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -230,4 +230,23 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
+          case IndividualRetrievedData(_, _, _, _, _, _, Some(saStatus)) =>
+            saStatus.status match {
+              case SAStatus.ReturnFound        => routes.TaxSituationController.confirmYourIncome()
+              case SAStatus.NoticeToFileIssued => routes.CheckYourAnswersController.checkYourAnswers()
+              case SAStatus.NoReturnFound      => routes.TaxSituationController.noReturnFoundExit()
+            }
+          case _: CompanyRetrievedData                                   => routes.CheckYourAnswersController.checkYourAnswers()
+          case _                                                         => routes.TechnicalExceptionController.technicalException()
```
AndyHWChung, line 248: this is impossible since there is only two things that `retrievedUserData` can be - we can remove this

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -230,4 +230,23 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
+        }
+    }) getOrElse routes.TechnicalExceptionController.technicalException()
```
AndyHWChung, line 250: without the `TechnicalExceptionController` we'd either have to do something like `sys.error` here or be a bit more lenient to say something like:
```scala
if(maybeTaxSituation.exists(isSA)) saStuff
else nonSaStuff
```
We've gone for the latter approach in other places in this service

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
AndyHWChung: to get the `sequence` working you'd need 
```scala
import cats.syntax.traverse._
``` 
as well

AndyHWChung: and you may need to be explicit about the type parameters:
```scala
.sequence[EitherT[Future,Error,*], SAStatusResponse]
```
since the first type parameter is looking for a first-order higher kinded type

meghana-viswanath: thanks @AndyHWChung, I'll update. Definitely right in keeping the next page logic all within the `JourneyService`

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
meghana-viswanath: Ah yes, ok, then should we persist the tax year in the session then? Fetch from session if present else calculate?

meghana-viswanath:
@AndyHWChung @priyankadhiman3006 refactored the tax situation api logic as per comments, looks much better Andy, thank you! The `EitherT` tripped me up a bit. Latest commit - https://github.com/hmrc/hec-applicant-frontend/pull/24/commits/c398350fc7a7b6e6bc0c6e85350f266fa7b4b10d

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
AndyHWChung: that'd ultimately be the safest thing to do I think - that means on the display tax situation page endpoint, we'd calculate the relevant tax year, then store it in session before actually displaying the page. There could be a `relevantTaxYear: Option[TaxYear]` somewhere in the session. That'd help us in the future when we come to handle that edge case we were talking about yesterday. 
If that's a lot of work to do now, we could get away with just calling `getTaxYear` here again and fix that up later when we actually come to work on that edge case later on. Don't want to massively increase the scope of the work here again when we're eventually gonna fix it up anyway.

meghana-viswanath: OK, makes sense to do as part of the edge case ticket. I'll put a note in there about it.

meghana-viswanath:
> Just for FYI :), if you want to get rid of sequence[EitherT[Future, Error, *] and map you can just do
> individualRetrievedData.sautr.traverse(taxCheckService.getSAStatus(_, taxYear))

Ooh nice! Let me do that!

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -230,4 +230,21 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
+          case TaxSituation.PAYE | TaxSituation.NotChargeable => routes.CheckYourAnswersController.checkYourAnswers()
+          case TaxSituation.SA | TaxSituation.SAPAYE          =>
+            saStatus.status match {
+              case SAStatus.ReturnFound        => routes.TaxSituationController.confirmYourIncome()
+              case SAStatus.NoticeToFileIssued => routes.CheckYourAnswersController.checkYourAnswers()
+              case SAStatus.NoReturnFound      => routes.TaxSituationController.noReturnFoundExit()
+            }
+        } getOrElse sys.error("Could not find tax situation in session")
+
+      case i: IndividualRetrievedData if i.sautr.isEmpty => routes.TaxSituationController.sautrNotFoundExit()
```
AndyHWChung, line 246: the presence of an SA UTR is only important if the user has selected a tax situation with SA in it - someone could select PAYE for instance and still be allowed to go to the CYA page. The check for this exit page should go just under 
```scala
case TaxSituation.SA | TaxSituation.SAPAYE =>
```
I think.

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -102,17 +134,38 @@ class TaxSituationController @Inject() (
...
+
+  // Placeholder for exit page when no tax return is found (HEC-987)
+  val noReturnFoundExit: Action[AnyContent] = authAction.andThen(sessionDataAction).async { implicit request =>
+    Ok(
+      s"Session is ${request.sessionData} back Url ::${journeyService.previous(routes.TaxSituationController.noReturnFoundExit())}"
+    )
+  }
+
+  // Placeholder for exit page when SAUTR not found (HEC-986)
+  val sautrNotFoundExit: Action[AnyContent] = authAction.andThen(sessionDataAction).async { implicit request =>
```
AndyHWChung, line 163: could we perhaps move these three endpoints (`sautrNotFoundExit`, `noReturnFoundExit`,  `confirmYourIncome`) to a separate controller like `SAController`? We're gonna have similar pages for CT which we could stick under something like `CTController` so that it's clear what the endpoints are for.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: Yeah, that makes sense, it's only applicable to users that have signed up for SA. I'll rework 👍

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
meghana-viswanath: yes, good idea 👍

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: I think we might have to go for the latter approach here too since after the rework based on this comment, some `updateAndNext` tests are failing. This is because we use the `taxSituationRoute` method not only to fetch the next page but also to understand if the next page is an exit page and in the latter case, we don't want to enforce presence of the tax situation answer.
I might have missed, but can't find an example of the approach suggested above, can you point me in the right direction @AndyHWChung ?

AndyHWChung: ah ok - I was thinking of something like this we have in the `JourneyService`:
```scala
private def entityTypeRoute(session: HECSession): Call = {
    val selectedEntityType = session.userAnswers.fold(_.entityType, _.entityType)
    val ggEntityType       = EntityType.fromRetrievedApplicantAnswers(session.retrievedUserData)

    if (selectedEntityType.contains(ggEntityType)) routes.TaxSituationController.taxSituation()
    else routes.EntityTypeController.wrongGGAccount()
  }
```
At the end there it's a bit more lenient where the check is `if(_.contains(..)) something else something else` - there's no hard check to see whether or not the `selectedEntityType` is really there

meghana-viswanath: ah ok, but where can we go if the tax situation answer is missing? Has to be an exit page like the technical exception page I had earlier?

AndyHWChung: in the `entityType` example, if the `entityType` is not currently there then it'd go to the `wrongGGAccount` page which technically isn't correct - the leniency there allows for some weird journeys, but it was based on a trust that that method would be called with the right state I suppose.

 To make it more water-tight, we could in theory make `next` the `entityType` page if it's not actually there which would make the user actually enter one in. In your ticket then that would mean making it go to the tax situation page. There's probably an argument to make it go to the `firstPage` (method in `JourneyService`) already to make the user go to the beginning to go through the journey properly without wiping any answers they might already have. The latter approach would be easier to maintain I think and it'd be nice to have that consistent approach in the other places like the `entityType` route.

AndyHWChung: but then actually, you'd end up on an infinite loop when doing the exit page check - perhaps for now we could take that lenient approach and maybe spend some more time thinking about if we want to make it more water-tight and how to do it in the future.

AndyHWChung: it might be that there's a better way to make that uplift to complete answers function work correctly, but not what alternatives there are atm. Might take some time to think about

meghana-viswanath: I'm not sure I fully understand the former approach. But with the latter approach of making it go to the `firstPage`, there's a possibility we would get stuck in an infinite loop when checking if we are at an exit page since we are working with the same session data?

meghana-viswanath: ah sorry, my page wasn't refreshed so just repeated what you had already said 🤦

AndyHWChung: yeah, we came to the same conclusion :laughing:  The former approach is basically the same as the latter one except we'd redirect to the page which captures the missing field - but then we'd end up in an infinite loop again.

AndyHWChung: this is a different scenario - the scenario in the ticket is talking about what if we get a 404 from the API call. Here, we wondering what to do if we're in the  `taxSituationRoute` in the `JourneyService` but there isn't actually a `taxSituation` answer in session - it shouldn't happen but we should think about the case where our service isn't working 100% as expected

meghana-viswanath: After huddle chat with @AndyHWChung and @priyankadhiman3006 about this, we've decided to unblock this PR by using the lenient approach and Andy will create a ticket and bring this up in the next tech refinement session.

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
meghana-viswanath: done in commit - https://github.com/hmrc/hec-applicant-frontend/pull/24/commits/aee388aae4dc9535ed0815f8cc6d7ee20810b504

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: done in commit - https://github.com/hmrc/hec-applicant-frontend/pull/24/commits/d667fc6070491757b46c8e37f1d81b1a1eff7eb7

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -230,4 +231,23 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
+
+    if (maybeTaxSituation.exists(ts => saTaxSituations.contains(ts))) {
+      session.retrievedUserData match {
+        case IndividualRetrievedData(_, _, Some(_), _, _, _, Some(saStatus)) =>
+          saStatus.status match {
+            case SAStatus.ReturnFound        => routes.SAController.confirmYourIncome()
+            case SAStatus.NoticeToFileIssued => routes.CheckYourAnswersController.checkYourAnswers()
+            case SAStatus.NoReturnFound      => routes.SAController.noReturnFoundExit()
+          }
+        case i: IndividualRetrievedData if i.sautr.isEmpty                   => routes.SAController.sautrNotFoundExit()
```
AndyHWChung, line 245: we'd get a match exception here is there was an SAUTR but there was no `saStatus` - we'd handle it properly in the other ticket though where we look at the missing field logic, but it may break the current looping logic in the uplift function depending on the state. Might be safer for now to have a separate case going to somewhere (not sure where) or to not have `if i.sautr.isEmpty` here for now.

meghana-viswanath: yeah, annoying that the compiler doesn't catch this, I guess if we did a  pattern match instead of an if guard, the compiler would complain. Might be better to do it that way even if it means a bit of ugliness around the pattern match since there are quite a few fields in the case class

meghana-viswanath: I'm not sure what to do here, we could go to the "return not found" page, but that's a bit misleading. If we leave it out completely then it won't pass QA? Can we just flag and let QA know in that case?

AndyHWChung: it'd probably wouldn't get noticed cos it's one of these theoretical errors which we can't trigger if we go through the FE properly. If we miss out `if i.sautr.isEmpty` from that guard and then stick a  `TODO` there to handle it properly then it should get picked up in that ticket where we handle missing fields better if we remember. When that ticket is done there'd be a better mechanism to handle that particular error.

AndyHWChung (approved):

priyankadhiman3006 (approved):

PR #27, opened 2021-09-10
HEC-986 SAUTR not found page

meghana-viswanath:
Details in ticket - https://jira.tools.tax.service.gov.uk/browse/HEC-986

### DIFF
--- a/test/uk/gov/hmrc/hecapplicantfrontend/controllers/SAControllerSpec.scala
+++ b/test/uk/gov/hmrc/hecapplicantfrontend/controllers/SAControllerSpec.scala
@@ -18,6 +18,7 @@ package uk.gov.hmrc.hecapplicantfrontend.controllers
 
 import java.time.LocalDate
 
+import org.jsoup.nodes.Document
 import play.api.inject.bind
 import play.api.mvc.Result
 import play.api.test.FakeRequest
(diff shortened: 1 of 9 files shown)


test/uk/gov/hmrc/hecapplicantfrontend/controllers/SAControllerSpec.scala
```diff
@@ -0,0 +1,113 @@
...
+
+        checkPageIsDisplayed(
+          performAction(),
+          messageFromMessageKey("sautrNotFound.title"),
+          doc => {
+            doc.select("#back").attr("href") shouldBe mockPreviousCall.url
+
+            testLink(
+              doc,
+              selector = "a:contains(go back and change your answer)",
```
AndyHWChung, line 89: the selectors here may make the test brittle in that it may break if the content ever changes on those links.  Does something like `"a.govuk-link"` select all of the links that you're interested in or does it bring in a whole load you don't want as well? Maybe something like `"a.govuk-link > p.govuk-body"` might filter out the ones you really want

meghana-viswanath: yeah, you are right, updated in commit - https://github.com/hmrc/hec-applicant-frontend/pull/27/commits/ea14b2713b43218a00d50a32c558641ce2382a4b

used a selector that you've suggested + with the href attr - `s"a.govuk-link[href=$url]"`

AndyHWChung (approved):

priyankadhiman3006 (approved):

nubz (approved):
Looks good to me

PR #28, opened 2021-09-10
HEC-987 No SA return found page

meghana-viswanath:
Details in ticket - https://jira.tools.tax.service.gov.uk/browse/HEC-987

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/views/NoReturnFound.scala.html
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/views/NoReturnFound.scala.html
@@ -11,6 +11,5 @@
 
 @layout(pageTitle = Some(title), backLocation = Some(back)) {
     <h1 class="govuk-heading-xl">@title</h1>
-    <p class="govuk-body">@Html(messages(s"$key.p1", appConfig.contactHmrcSa))</p>
-
+    <p class="govuk-body govuk-!-margin-bottom-9">@Html(messages(s"$key.p1", appConfig.contactHmrcSa))</p>
 }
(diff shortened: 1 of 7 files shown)


AndyHWChung (approved):

priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/views/NoReturnFound.scala.html
```diff
@@ -0,0 +1,16 @@
...
+
+@this(layout: Layout)
+
+@(back: Call)(implicit request: Request[_], messages: Messages, appConfig: AppConfig)
+@key = @{"noReturnFound"}
+@title = @{messages(s"$key.title")}
+
+@layout(pageTitle = Some(title), backLocation = Some(back)) {
+    <h1 class="govuk-heading-xl">@title</h1>
+    <p class="govuk-body">@Html(messages(s"$key.p1", appConfig.contactHmrcSa))</p>
```
nubz, line 14: can we add the extra css class `govuk-!-margin-bottom-9` to the final paragraph - this creates the space we need at the bottom of the page (because we have a get-technical-help link in our layout template) so it would be `<p class="govuk-body govuk-!-margin-bottom-9">@Html(...`

nubz: also think we need to check the contact hmrc link - this page is not exclusively for SA users - company users will also see it so wondering how we can handle that - maybe we can leave it until we do the company journeys...

meghana-viswanath: Thanks Ali! Added the space.

About the contact hmrc link, think we can do it when we get to the company journeys - we'd just check the applicant type and fetch the corresponding url and pass to the template

meghana-viswanath: Hi @nubz, are you happy with this?

nubz: yes indeed @meghana-viswanath - we can tackle it when we do company journeys

PR #29, opened 2021-09-13
HEC-985 Confirm your income page

meghana-viswanath:
Details in ticket - https://jira.tools.tax.service.gov.uk/browse/HEC-985

Note: I'll create a follow-on PR for calculating and storing the tax year in the session so it can be reused instead of calculating as and when required to avoid the edge case where the tax year might change from one page to the next


app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
```diff
@@ -33,6 +33,7 @@ object UserAnswers {
     licenceTimeTrading: Option[LicenceTimeTrading],
     licenceValidityPeriod: Option[LicenceValidityPeriod],
     taxSituation: Option[TaxSituation],
+    incomeConfirmation: Option[IncomeConfirmation],
```
AndyHWChung, line 36: Could we perhaps go for something like `saIncomeDeclared: SAIncomeDeclared` or something to reflect the SA nature of it (we'll probably have a similar but separate field for the CT journey) and to reflect the Boolean-type nature of the field?

app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
```diff
@@ -42,6 +43,7 @@ object UserAnswers {
     licenceTimeTrading: LicenceTimeTrading,
     licenceValidityPeriod: LicenceValidityPeriod,
     taxSituation: TaxSituation,
+    incomeConfirmation: IncomeConfirmation,
```
nubz, line 46: Should this not continue to be an Option? It is possible for no tax return to be found for a user and therefore for them never to be asked this question and also PAYE users do not answer this question

test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
```diff
@@ -572,6 +572,18 @@ class JourneyServiceSpec extends AnyWordSpec with Matchers with MockFactory with
           }
         }
 
+        "the confirm your income page" in {
```
AndyHWChung, line 575: it'd be useful to have `your SA income` here to help distinguish it from the CT version later on

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -165,6 +165,7 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
                 Some(licenceTimeTrading),
                 Some(licenceValidityPeriod),
                 Some(taxSituation),
+                Some(incomeConfirmation),
```
AndyHWChung, line 168: here since the income field should be optional in `CompleteUserAnswers` you'll need an extra case(s) to uplift when this field is populated and when the business rules dictate it should be, i.e. `taxSituation` is an SA type one and the SA status is `ReturnFound`. In other cases we don't care whether or not this field is populated or not

AndyHWChung: it might get a bit messy though so you we might want to do something like:
```scala
private def allAnswersComplete(incompleteAnswers: IncompleteUserAnswers, session: HECSession): Boolean = ???
...
private def upliftToCompleteAnswersIfComplete(session: HECSession, current: Call): Either[Error, HECSession] = {
...
val updatedSession = session.userAnswers match {
        case _ if isExitPageNext => 
               // session
        case incomplete: IncompleteUserAnswers if allAnswersComplete(incomplete) =>
             // fill in  CompleteUserAnswers(...)
        case _ => session
}
```
where all the business logic is in that new `allAnswersComplete` method instead of having a load of cases in the existing uplift method

app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
meghana-viswanath: very true, my mistake, will update

app/uk/gov/hmrc/hecapplicantfrontend/controllers/SAController.scala
```diff
@@ -57,3 +107,17 @@ class SAController @Inject() (
...
+
+object SAController {
+
+  val incomeConfirmationOptions: List[IncomeConfirmationOption] =
+    IncomeConfirmation.values.map(IncomeConfirmationOption.incomeConfirmationOption)
+
+  def incomeConfirmationForm(options: List[IncomeConfirmation]): Form[IncomeConfirmation] =
+    Form(
+      mapping(
+        "confirmYourIncome" -> of(FormUtils.radioFormFormatter(options))
```
priyankadhiman3006, line 119: Instead of confirmYourIncome can it e changed to SAIncome?

app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
meghana-viswanath: yes, I like that better 👍

test/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationControllerSpec.scala
```diff
@@ -179,6 +179,7 @@ class TaxSituationControllerSpec
                 LicenceTimeTrading.TwoToFourYears,
                 LicenceValidityPeriod.UpToThreeYears,
                 TaxSituation.PAYE,
+                IncomeConfirmation.Yes,
```
priyankadhiman3006, line 182: I believe, the income confirmation page will appear only if the taxSitution is SA/ SAPAYE, if TaxSituation.PAYE, then IncomeConfirmation.Yes can never happen ?

test/uk/gov/hmrc/hecapplicantfrontend/controllers/SAControllerSpec.scala
```diff
@@ -126,6 +130,211 @@ class SAControllerSpec
...
+            checkIsRedirect(performAction("confirmYourIncome" -> "1"), mockNextCall)
+          }
+
+          "the user has previously completed answering questions" in {
+            val answers        = CompleteUserAnswers(
+              LicenceType.DriverOfTaxisAndPrivateHires,
+              LicenceExpiryDate(TimeUtils.today().minusDays(10L)),
+              LicenceTimeTrading.ZeroToTwoYears,
+              LicenceValidityPeriod.UpToOneYear,
+              TaxSituation.PAYE,
```
priyankadhiman3006, line 312: We should use TaxSituation.SA/ SAPAYE if we want IncomeConfirmation.Yes

test/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationControllerSpec.scala
```diff
@@ -179,6 +179,7 @@ class TaxSituationControllerSpec
                 LicenceTimeTrading.TwoToFourYears,
                 LicenceValidityPeriod.UpToThreeYears,
                 TaxSituation.PAYE,
```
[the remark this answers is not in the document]
meghana-viswanath: you are right, @nubz pointed this out too, I'll update

test/uk/gov/hmrc/hecapplicantfrontend/controllers/CheckYourAnswersControllerSpec.scala
```diff
@@ -98,6 +98,7 @@ class CheckYourAnswersControllerSpec
           LicenceTimeTrading.ZeroToTwoYears,
           LicenceValidityPeriod.UpToTwoYears,
           TaxSituation.PAYE,
```
priyankadhiman3006, line 100: same here

app/uk/gov/hmrc/hecapplicantfrontend/controllers/SAController.scala
meghana-viswanath: I think we might just end up using the same page for CT too, so a more generic name would suit better I think

meghana-viswanath: Actually, looking at the content tracker, the messaging is a bit different, so it might be different page, I'll update to reflect that this is a SA specific page 👍

app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
meghana-viswanath: I'm thinking we could reuse the sealed trait for both SA and CT, so maybe calling it something generic like `IncomeDeclaration`? How does that sound?

AndyHWChung: I think reusing the case class would work for both - it's not likely that a change to one would not affect the other. `Yes` and `No` sound like a Boolean type to me so `IncomeDeclared` sounds a bit better to me, or `IncomeDeclaration` extended by `IncomeDeclared` and `IncomeNotDeclared` would make it clear as well. Not sure what is being answered by yes and no with `IncomeDeclaration`.

meghana-viswanath: yeah, `IncomeDeclared` sounds good 👍

meghana-viswanath:
@AndyHWChung @priyankadhiman3006 @nubz I've updated the PR to make all changes as per comments (thanks!). But since there is quite a lot of renaming might be better to look at all the changes together even though I've done new commits on top of the old.

priyankadhiman3006:
build has failed

meghana-viswanath:
> It seems build has failed due to some import
> import uk.gov.hmrc.hecapplicantfrontend.models.UserAnswers.{CompleteUserAnswers, IncompleteUserAnswers}

Just fixed it 🤞

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
```diff
@@ -99,8 +99,11 @@ class TaxSituationController @Inject() (
             maybeSaStatus <- fetchSAStatus(individualRetrievedData, taxSituation)
 
             updatedRetrievedData = individualRetrievedData.copy(saStatus = maybeSaStatus)
-            updatedAnswers       =
-              request.sessionData.userAnswers.unset(_.taxSituation).copy(taxSituation = Some(taxSituation))
+            // wipe the SA income declared answer since it is dependent on the tax situation answer
+            updatedAnswers       = request.sessionData.userAnswers
+                                     .unset(_.taxSituation)
+                                     .unset(_.saIncomeDeclared)
```
AndyHWChung, line 105: in this scenario:
- user selecs one of the SA options
-  user answers the sa income declaration question
- user gets back on to this tax situation page
- user doesn't change answer and submits

we don't want to wipe the answer to the income question cos it hasn't changed

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -268,3 +249,61 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
+object JourneyServiceImpl {
+  def licenceTypeForIndividualAndCompany(licenceType: LicenceType): Boolean = licenceType match {
+    case LicenceType.DriverOfTaxisAndPrivateHires => false
+    case _                                        => true
+  }
+
+  /**
+    * Expect the entity type to be specified only for individual or company licence types
+    */
+  private def checkEntityType(licenceType: LicenceType, entityType: Option[EntityType]): Boolean =
```
AndyHWChung, line 262: maybe the name of the method here can be a bit more specific, like `checkEntityTypePresentIfRequired`? That name is a bit long though

priyankadhiman3006:
now you have to rebase 🙈 because i merged the expiry date changes  apologies 🙏

app/uk/gov/hmrc/hecapplicantfrontend/controllers/SAController.scala
```diff
@@ -16,34 +16,84 @@
...
+    val saIncomeDeclared = request.sessionData.userAnswers.fold(_.saIncomeDeclared, _.saIncomeDeclared)
+    val form = {
+      val emptyForm = SAController.saIncomeDeclarationForm(IncomeDeclared.values)
+      saIncomeDeclared.fold(emptyForm)(emptyForm.fill)
+    }
+    Ok(saIncomeStatementPage(form, back, SAController.incomeDeclaredOptions, getTaxYear(timeProvider.currentDate)))
+  }
+
+  val saIncomeStatementSubmit: Action[AnyContent] = authAction.andThen(sessionDataAction).async { implicit request =>
+    def handleValidAnswer(incomeDeclared: IncomeDeclared): Future[Result] = {
```
priyankadhiman3006, line 64: Is there any invalid Answer we are handling or we need to rename this method?

test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
```diff
@@ -1199,6 +1304,172 @@ class JourneyServiceSpec extends AnyWordSpec with Matchers with MockFactory with
 
     }
 
+    "JourneyServiceImpl.allAnswersComplete" must {
```
AndyHWChung, line 1307: it might be safer to test what this method does indirectly under the `not convert incomplete answers to complete answers when all questions have been answered and` section in the spec - in theory, we could change the code to not call `allAnswersComplete` and change the behaviour of `updateAndNext` and these specs would still pass - it's really the `updateAndNext` behaviour which we're most interested in

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
meghana-viswanath: if the user goes back to the tax situation page, they have to then continue on to the income declaration page again before submitting, so they will have to answer the income question again even if they decide to not change the tax situation answer

app/uk/gov/hmrc/hecapplicantfrontend/controllers/SAController.scala
AndyHWChung: I think that may have come from the other parts of the code where it's named in a similar way - it's being called just after the form validation takes place

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
meghana-viswanath: in the scenario where user does change the tax situation answer, they should be allowed to update the income declaration too I am guessing?

app/uk/gov/hmrc/hecapplicantfrontend/controllers/SAController.scala
meghana-viswanath: the first arg to the fold handles an invalid answer - 
```
SAController
      .saIncomeDeclarationForm(IncomeDeclared.values)
      .bindFromRequest()
      .fold(
        formWithErrors =>
          Ok(
            saIncomeStatementPage(
              formWithErrors,
              journeyService.previous(routes.SAController.saIncomeStatement()),
              SAController.incomeDeclaredOptions,
              getTaxYear(timeProvider.currentDate)
            )
          ),
        handleValidAnswer
      )
```

test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
meghana-viswanath: My thinking was that it would be easier to review and reason about the various cases (since there are quite a few and we'll likely add more with the company journeys) if we are looking at a smaller unit of code and that the actual thing we are testing would get a bit lost if we test it indirectly.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -268,3 +249,61 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
+  /**
+    * Expect the SA income to be declared only for an SA tax situation whose return has been found
+    */
+  private def checkSAIncomeDeclared(
+    taxSituation: TaxSituation,
+    saIncomeDeclared: Option[IncomeDeclared],
+    retrievedUserData: RetrievedApplicantData
+  ): Boolean =
+    (retrievedUserData, saIncomeDeclared) match {
+      case (IndividualRetrievedData(_, _, _, _, _, _, Some(SAStatusResponse(_, _, ReturnFound))), None)
```
priyankadhiman3006, line 278: Just want to confirm: Are we checking here that the Individual is logged in with Return Found status , but if the tax situation is selected as SA/ SAPAYE, then it means SAIncome Not declared?
If yes, then the comment written upon is misleading or it appears to me

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: I can look at renaming, didn't bother too much since it was private and a comment was in place

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: sorry @priyankadhiman3006, it's a disaster of a pattern match really, but couldn't make it more readable and succinct at the same time.

that particular case means that 
- the user is an individual
- with tax situation SA/SAPAYE
- with a `ReturnFound` status 
- *but* is missing the SA declared income answer

In this scenario answers are not complete, in all other cases, they are complete

test/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswersSpec.scala
```diff
@@ -47,6 +47,7 @@ class UserAnswersSpec extends AnyWordSpec with Matchers {
           LicenceTimeTrading.TwoToFourYears,
           LicenceValidityPeriod.UpToFiveYears,
           TaxSituation.PAYE,
+          Some(IncomeDeclared.Yes),
```
priyankadhiman3006, line 50: I thinkIncomeDeclared page will not appear  if TaxSituation.PAYE,, in order to use Some(IncomeDeclared.Yes),, i think you should use TaxSituation.SA?

priyankadhiman3006: same for others  down below

meghana-viswanath: I can change but really the actual values here shouldn't affect that tests since they aren't checking that the answers fit together as expected

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
AndyHWChung: it might just be me, but my eyes are trained to look at the function signature first with the name to look at what the function is doing - in most cases that should be enough to not require a comment which takes a bit of extra effort to read/maintain. It wouldn't harm to have a more descriptive name at least :slightly_smiling_face:

test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
AndyHWChung: that makes sense - I think there's still a risk that we lose some coverage of the actual behaviour of the "public" endpionts provided by the service. Those are the ones which are most important to us cos they'd be used by other things in the whole service itself. We could still somehow manufacture a situation where the tests for `allAnswersComplete` are all passing but `updateAndNext` is not behaving as we want say if, for example, `updateAndNext` no longer calls `allAnswersComplete` or behaviour is added to `updateAndNext` which overrides some of the behaviour in `allAnswersComplete`

test/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswersSpec.scala
priyankadhiman3006: yes, but it still is wrong :)

meghana-viswanath: it's most definitely wrong, but in the context of what the test is doing and since the model allows for it, it's like using an invalid SAUTR - we know in the real world it'll fail, but in the context of the test, it's fine

i'll update just to not cause any confusion 👍

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: definitely no harm, my whole intention of moving it to a separate function was so it becomes easy to follow using function names, so makes sense to make them as descriptive as possible 👍

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
AndyHWChung: if they've already answered the income declaration before and the tax situation answer hasn't changed there doesn't seem to be a reason for them to re-answer the the income declaration question again. They'd get to cya gain and then they'd be able to change the income declaration answer from there if they want to. Best to check with David perhaps though to get the actual rules.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: fixed in commit - https://github.com/hmrc/hec-applicant-frontend/pull/29/commits/e5647be052232498df138e1bf3a9523d7f3d6692

test/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswersSpec.scala
meghana-viswanath: fixed in commit - https://github.com/hmrc/hec-applicant-frontend/pull/29/commits/e5647be052232498df138e1bf3a9523d7f3d6692

AndyHWChung: yeah, a lot of parts in our service are dumb to what's going on in other parts of the service. There are arguments on both sides on how the test data should reflect that I think - on one hand, having it reflect the actual form of the data we expect shows our intentions more clearly but having it not do so can make it more obvious exactly what parts of the data the thing your testing cares about. With the latter approach at least we don't have to think too hard about how to set up the test data and not have to worry about updating it to the "correct" thing if rules change in the future.

conf/messages
```diff
@@ -216,4 +216,12 @@ sautrNotFound.problem.p1 = You’ll need to <a href="{0}" class="govuk-link">con
 
 # No Return Found
 noReturnFound.title = You cannot do a tax check at the moment
-noReturnFound.p1 = You will need to <a href="{0}" class="govuk-link">contact HMRC</a> before doing your tax check.
\ No newline at end of file
+noReturnFound.p1 = You will need to <a href="{0}" class="govuk-link">contact HMRC</a> before doing your tax check.
+
+# Confirm Your Income
+confirmYourIncome.title = Did you include income you earned using this licence in your tax return?
```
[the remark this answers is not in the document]
meghana-viswanath: the [prototype](https://hec-applicant-tax-check.herokuapp.com/v5_3/confirm-your-income) I've got says the same thing. where did you see this title?

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala
meghana-viswanath: waiting on David 👍

test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
meghana-viswanath: yeah, definitely agree that this service is super important and used practically everywhere else in the code so making it more robust even if it means repeating a few tests by indirectly testing them will give us more confidence. I don't have a strong argument for having fewer tests 😂, so I'll add 👍 

I did add a few tests [here](https://github.com/hmrc/hec-applicant-frontend/blob/9493ef9457adb89bfba6f5953d45e1898e09ded2/test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala#L689), but doesn't cover all cases

meghana-viswanath: added the missing test case - https://github.com/hmrc/hec-applicant-frontend/pull/29/commits/c82427eaaa1cbb26c2432b3a32383de9c59e9119

conf/messages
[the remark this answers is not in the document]
priyankadhiman3006: I saw it yesterday, haven't checked today :)

test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
AndyHWChung: :+1:  thank you :slightly_smiling_face:

nubz (approved):
Front end looks good to me

AndyHWChung (approved):

PR #31, opened 2021-09-14
HEC-872 timeout handling

nubz:


nubz:
I put a Do Not Merge label on it until we agree it can be pulled into this sprint... however reviews welcome!

app/uk/gov/hmrc/hecapplicantfrontend/config/AppConfig.scala
```diff
@@ -54,6 +55,9 @@ class AppConfig @Inject() (config: Configuration, contactFrontendConfig: Contact
 
   lazy val signOutUri: String = config.get[String]("auth.sign-out.uri")
 
+  val authTimeoutSeconds: Int          = config.get[Duration]("auth.gg.timeout").toSeconds.toInt
```
AndyHWChung, line 58: `.get[FiniteDuration]` here would help protect against the unlikely scenario that someone tries to modify it to an infinite timeout

conf/application.conf
```diff
@@ -148,6 +148,10 @@ auth {
 
   sign-out.uri = "http://localhost:9949/auth-login-stub/session/logout"
 
+  gg.timeout = 900 seconds
```
AndyHWChung, line 151:
```
sign-out {
  uri = ...
  timeout = ...
  countdown ...
}
```
seems to make a bit more sense as the new properties are kind of properties of signing out rather than of gg in general.

AndyHWChung: also may something like `inactivity-timeout` and `inactivity-countdown` would make it more obvious at first glance what these config params are for

app/uk/gov/hmrc/hecapplicantfrontend/config/AppConfig.scala
nubz: done

conf/application.conf
nubz: done

app/uk/gov/hmrc/hecapplicantfrontend/controllers/SignOutController.scala
```diff
@@ -0,0 +1,37 @@
...
+@Singleton
+class SignOutController @Inject() (
+  mcc: MessagesControllerComponents,
+  timedOutPage: TimedOut
+) extends FrontendController(mcc)
+    with I18nSupport
+    with Logging {
+
+  def signOutFromTimeout: Action[AnyContent] = Action.async { implicit request =>
+    Ok(timedOutPage()).withNewSession
```
AndyHWChung, line 35: worth a unit test here - do you need help with that? I'd be happy to do one for you here

nubz: Yes, a new SignOutControllerSpec, we may end up adding more to it if we decide on a regular signed out page too, not sure how to assert an existing session has been destroyed which should probably be part of the unit tests... so please do set it up @AndyHWChung thanks

AndyHWChung: spec now added

priyankadhiman3006 (approved):

meghana-viswanath (approved):

PR #36, opened 2021-09-21
HEC-1058 json library swap

meghana-viswanath:


app/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckData.scala
```diff
@@ -16,28 +16,56 @@
...
     taxDetails: CompanyTaxDetails
-  ) extends HECTaxCheckData
+  ) extends HECTaxCheckData {
+    val entityType: EntityType = EntityType.Company
+  }
+
+  implicit val formatIndividual: OFormat[IndividualHECTaxCheckData] = Json.format[IndividualHECTaxCheckData]
+  implicit val formatCompany: OFormat[CompanyHECTaxCheckData]       = Json.format[CompanyHECTaxCheckData]
+
+  implicit val format: OFormat[HECTaxCheckData] = new OFormat[HECTaxCheckData] {
```
AndyHWChung, line 49: I wonder if we can get away with just having `Jsonx.formatSealed` here - the fields within the nested case classes may be enough to be able to distinguish between `IndividualHECTaxCheckData` and `CompanyHECTaxCheckData` as they were before. To make sure that's guaranteed we could change the field names if we really wanted to like:
```scala
final case class IndividualHECTaxCheckData(
    individualApplicantDetails: IndividualApplicantDetails,
    licenceDetails: LicenceDetails,
    individualTaxDetails: IndividualTaxDetails
  )
```
That's slightly unclean though. The same thing applies to the other places where we've got a new `type` field.

app/uk/gov/hmrc/hecapplicantfrontend/models/CTStatus.scala
```diff
@@ -31,6 +33,14 @@ object CTStatus {
 
   case object NoReturnFound extends CTStatus
 
-  implicit val format: OFormat[CTStatus] = derived.oformat()
+  @SuppressWarnings(
+    Array(
```
AndyHWChung, line 37: this is slightly annoying :disappointed: Suppressing might introduce risk of bad things seeping into the code but having to keep track of these here is not ideal. A `@silent` annotation here would be nicer I think to avoid having too much of this type of thing across the code.

meghana-viswanath: how exactly do I use it? just putting a `@silent` above or at the start of the line doesn't seem to be doing it. 
do I need to enable the plugin?

AndyHWChung: I thought it'd just work by sticking it just above the line like this:
https://github.com/hmrc/cgt-property-disposals/blob/d988fad9f0406678407064c46b948ff7114ea428/app/uk/gov/hmrc/cgtpropertydisposals/models/returns/ExampleCompanyDetailsAnswers.scala#L42

I think the plugin is already being pulled in with the `silencer-plugin` part in `build.sbt`. I'll try to double check to see if anything special is needed

app/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckData.scala
meghana-viswanath: exactly my thoughts too, cool, I'll update 👍

meghana-viswanath: not sure this is applicable to the other two cases - `UserAnswers` and `RetrievedApplicantData`.  

For `UserAnswers`, we could prefix the fields with `maybe` for `IncompleteAnswers` but not ideal.

For `RetrievedApplicantData` the only difference is the presence of the mandatory `dateOfBirth` field in `IndividualRetrievedData`. But if that changes, we might have issues.

AndyHWChung: oh yeah - for `UserAnswers` then we may have no choice cos the idea is that the fields mirror each other in the incomplete versions and the complete versions. Having `maybe` in the name is a bit annoying but the `type` field just makes it a bit more invisible when we're actually using the case class in the code. It might also look a bit odd with something prefixed with `maybe` in the `IncompleteAnswers` and the same field not having it in the `CompleteAnswers` but where the field in the `CompleteAnswers` is actually optional.

Having thought about it, I'm thinking what you had originally might have been the safest thing after all - I was just trying to save us writing a lot of code to support it. Maybe ignore my original comment then - sorry!

app/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckData.scala
```diff
@@ -16,28 +16,56 @@
...
+
+    override def reads(json: JsValue): JsResult[HECTaxCheckData] = {
+      val entityType = (json \ "type")
+        .validate[EntityType]
+        .getOrElse(sys.error("Invalid HECTaxCheckData type"))
```
AndyHWChung, line 54: I think you can just `flatMap` on the `JsResult` from `validate` here and map onto the entity type instead of doing `getOrElse`.

app/uk/gov/hmrc/hecapplicantfrontend/models/CTStatus.scala
meghana-viswanath: looks like there is support for this in scala 2.12.13 (which is the version we are on) using a `@nowarn` annotation - https://github.com/scala/scala/pull/8373

still trying to figure it out

app/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckCode.scala
```diff
@@ -16,14 +16,12 @@
...
-import play.api.libs.json.Format
+import play.api.libs.json.{Format, Json}
 
 final case class HECTaxCheckCode(value: String) extends AnyVal
 
 object HECTaxCheckCode {
 
-  implicit val format: Format[HECTaxCheckCode] =
-    implicitly[Format[String]].inmap(HECTaxCheckCode(_), _.value)
+  implicit val format: Format[HECTaxCheckCode] = Json.valueFormat[HECTaxCheckCode]
```
priyankadhiman3006, line 25: Not a big issues but i think [HECTaxCheckCode] after valueFormat is optional if we are giving return type like( format: Format[HECTaxCheckCode])

app/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckData.scala
meghana-viswanath: 👍 yeah, unfortunately this is the only uniform approach AFAIK that can be applied to all cases

app/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckCode.scala
meghana-viswanath: yeah, don't think it's big issue, might just keep as is if that's ok

app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
```diff
@@ -80,7 +105,29 @@ object UserAnswers {
...
+        .validate[UserAnswersType]
+        .getOrElse(sys.error("Invalid UserAnswers type"))
+
+      userAnswersType match {
+        case UserAnswersType.Incomplete => formatIncomplete.reads(json)
+        case UserAnswersType.Complete   => formatComplete.reads(json)
+      }
+    }
+
+    override def writes(o: UserAnswers): JsObject = {
```
priyankadhiman3006, line 123: We are using this at so many places, any chance we can generalise it and put in some utils so as to reuse it? something like writes[T](o: T): JsObject?
same for reads?

app/uk/gov/hmrc/hecapplicantfrontend/models/CTStatus.scala
meghana-viswanath: the other option is to use `@SuppressWarnings(Array("org.wartremover.warts.All"))`

app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
meghana-viswanath: I don't think it's easy to write a type-parameterised function for this since each implementation is very specific to the the parameter's sub classes.

app/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckData.scala
meghana-viswanath: updates in commit - https://github.com/hmrc/hec-applicant-frontend/pull/36/commits/ef7394cca93599546edbca09593cdf9933cfa25f

priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/models/CTStatus.scala
AndyHWChung: that's probably the easiest thing to go for for now yeah, let's just use that

meghana-viswanath: updated - https://github.com/hmrc/hec-applicant-frontend/pull/36/commits/66bec46c492c3896f50dabeb313398f00057cf62

meghana-viswanath:
@AndyHWChung @priyankadhiman3006 I've added tests for the readers/writes in [this commit](https://github.com/hmrc/hec-applicant-frontend/pull/36/commits/9587e373196df29337317e9de2787861a707e09f), can you review them, thanks

test/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckDataSpec.scala
```diff
@@ -0,0 +1,126 @@
...
+        )
+
+      val individualJson = Json.parse("""{
+          | "applicantDetails":{
+          |    "ggCredId":"ggCredId",
+          |    "name":{
+          |      "firstName":"first",
+          |      "lastName":"last"
+          |    },
+          |    "dateOfBirth":"20210922"
```
AndyHWChung, line 62: this test and a similar one in `RetrievedApplicantDataSpec` will only pass today - it'd be more reliable to hard code some random date for the date of birth so it stays constant through time

priyankadhiman3006:
Just a question, are we going to merge this now or will do the similar changes in other repos and then merge on the same day?
Cause if this merged and others may break when we try to validate tax code in LB.

meghana-viswanath:
> Just a question, are we going to merge this now or will do the similar changes in other repos and then merge on the same day?
> Cause if this merged and others may break when we try to validate tax code in LB.

Good question, I think it's better to do them all together @AndyHWChung ?

priyankadhiman3006 (approved):

test/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckDataSpec.scala
meghana-viswanath: good spot!

AndyHWChung:
> Just a question, are we going to merge this now or will do the similar changes in other repos and then merge on the same day?
> Cause if this merged and others may break when we try to validate tax code in LB.

Yeah, we should merge everything together including updates to `hec-ui-journey-tests` where it's calling our BE test-only endpoint. It'd be a good idea just to clear our mongo databases in the test environments as well once it's merged in

meghana-viswanath:
> > Just a question, are we going to merge this now or will do the similar changes in other repos and then merge on the same day?
> > Cause if this merged and others may break when we try to validate tax code in LB.
> 
> Yeah, we should merge everything together including updates to `hec-ui-journey-tests` where it's calling our BE test-only endpoint. It'd be a good idea just to clear our mongo databases in the test environments as well once it's merged in

I'll keep this in `Draft` until we are ready to merge

priyankadhiman3006:
Then again, other code will keep on piling  and you have to do rebase and stuff. i think we should divide and conquer :) happy to help

test/uk/gov/hmrc/hecapplicantfrontend/models/HECTaxCheckDataSpec.scala
meghana-viswanath: updated - https://github.com/hmrc/hec-applicant-frontend/pull/36/commits/516e39e1f2ed214fd927899352fcefc5e18f86ef

AndyHWChung (approved):

PR #38, opened 2021-09-24
Hec 1045 crn

priyankadhiman3006:
ticket -> https://jira.tools.tax.service.gov.uk/browse/HEC-1045
This Pr contains address the following changes :
1) company registration page, and navigation to company found page and company not found page
2) tax Situation made optional, since it's not in Company  journey. Changes in some html files because of this change
3) test cases related to CRN and tax situation changes


conf/messages
```diff
@@ -212,4 +213,12 @@ hmrcTimeoutDialog.title = You’re about to be signed out
...
\ No newline at end of file
+timedOut.signIn = Sign in
+
+# CRN
+crn.title = What is your company registration number?
+crn.hint = Companies House sent this to you when you set up your company. This is also called a company number. For example, SN898989 or 12345678.
+crn.link = Search for a company registration number
+crn.error.required = Enter your Company Registration Number
+crn.error.nonAlphanumericChars = The applicant’s company registration number must only include letters A to Z and numbers 0 to 9
+crn.error.crnInvalid = Enter the applicant's company registration number in the correct format
```
nubz, line 224: these error messages don't match what is in the content tracker v6 row 9 for this field

priyankadhiman3006: I think the content racker messages are old. David has mentioned the new one in ticket description. I will double check from david

priyankadhiman3006: "Your company registration number must be 10 characters long" this message is however wrong, cause as per new rule CRN can be 7/8 chars, so i assume the ticket one is correct , but again will double check :)

nubz: yes - I have pointed that out on the error messages ticket I have just been doing on the prototype

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
```diff
@@ -16,27 +16,115 @@
...
+              )
+            ),
+          handleValidCrn
+        )
+
+    }
+
+  private def checkCompanyName(crn: CRN)(implicit request: RequestWithSessionData[_]): EitherT[Future, Error, Call] =
+    for {
+      companyHouseDetailsOpt <- companyDetailsService.findCompany(crn)
```
AndyHWChung, line 80: if the submitted CRN is the same as the one in session and there's already a name for it, there's an optimisation we could make here to not call the companies house API again and just keep the session as is

app/uk/gov/hmrc/hecapplicantfrontend/services/CompanyDetailsService.scala
```diff
@@ -44,14 +44,15 @@ class CompanyDetailsServiceImpl @Inject() (
...
-  )(implicit hc: HeaderCarrier): EitherT[Future, Error, CompanyHouseDetails] =
+  )(implicit hc: HeaderCarrier): EitherT[Future, Error, Option[CompanyHouseDetails]] =
     companyDetailsConnector.findCompany(companyNumber).subflatMap { httpResponse =>
-      if (httpResponse.status =!= OK)
-        Left(Error(s"Response to get company details came back with status ${httpResponse.status}"))
-      else
+      if (httpResponse.status === OK) {
         httpResponse
-          .parseJSON[CompanyHouseDetails]
+          .parseJSON[Option[CompanyHouseDetails]]
```
AndyHWChung, line 51: should this be something like:
```scala
httpResponse
  .parseJSON[CompanyHouseDetails]
  .bimap(Error(_), Some(_))
```
? We're generally expecting the JSON to be there if we get a 200 I think.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -159,16 +160,20 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
                 Some(licenceValidityPeriod),
                 Some(taxSituation),
```
AndyHWChung, line 161: `Some(taxSituation)` now needs to be just `taxSitation` to allow for the cases where it might not be there for companies. Then in `allAnswersComplete` below we need a check that `taxsSituation` is populated if the user is an individual

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
priyankadhiman3006: yeah, i can put that check

app/uk/gov/hmrc/hecapplicantfrontend/services/CompanyDetailsService.scala
priyankadhiman3006: yes, will make that change

app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
```diff
@@ -47,7 +48,9 @@ object UserAnswers {
     licenceValidityPeriod: Option[LicenceValidityPeriod],
     taxSituation: Option[TaxSituation],
     saIncomeDeclared: Option[IncomeDeclared],
-    entityType: Option[EntityType]
+    entityType: Option[EntityType],
+    crn: Option[CRN],
+    companyName: Option[CompanyHouseName]
```
meghana-viswanath, line 53: should `companyName` be added to retrieved user data? don't think it's a user answer

app/uk/gov/hmrc/hecapplicantfrontend/models/CompanyHouseDetails.scala
```diff
@@ -18,11 +18,21 @@ package uk.gov.hmrc.hecapplicantfrontend.models
...
   implicit val config: Aux[Json.MacroOptions]           = JsonConfiguration(SnakeCase)
   implicit val userFormat: OFormat[CompanyHouseDetails] = Json.format[CompanyHouseDetails]
+
+  implicit def optionFormat[T : Format]: Format[Option[T]] = new Format[Option[T]] {
+    override def reads(json: JsValue): JsResult[Option[T]] = json.validateOpt[T]
+
+    override def writes(o: Option[T]): JsValue = o match {
+      case Some(t) ⇒ implicitly[Writes[T]].writes(t)
+      case None ⇒ JsNull
+    }
```
meghana-viswanath, line 35: where is this used? also, it looks like a generic function and should probably be defined in a utils object

app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
priyankadhiman3006: @AndyHWChung sholud we include company Name under company retrieve data?

app/uk/gov/hmrc/hecapplicantfrontend/models/CompanyHouseDetails.scala
priyankadhiman3006: I added this to format the httpResponse to Option[CompanyhouseDetails], but we need a value in case of 200, so will change as suggested by Andy (above)

app/uk/gov/hmrc/hecapplicantfrontend/models/UserAnswers.scala
AndyHWChung: I think there's a case to put it there yeah - it's been retrieved by us in the BE rather than the user inputting it

priyankadhiman3006: I was thinking about it and what if the user go back enter different CRN , then company name changes , shouldn't it be better to put in a session? If we try to store it in company Retrieved case class , then we had to have a fold and unset method for those case classes as well and i not going to be easy?
I think RetrievedApplicantData hold the values , which we know are fixed and not going to change till the end of the session. What do you guys think?

AndyHWChung: if it's in the `RetrievedApplicantData` then we should be able to change it easily still without `unset` methods like we have for `saStatus` here in the `TaxSituationController`:
https://github.com/hmrc/hec-applicant-frontend/blob/master/app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxSituationController.scala#L98
The `saStatus` field is similar to what we have here I think - it's something we retrieve about the user rather than something the user tells us.

priyankadhiman3006: updated

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
priyankadhiman3006: updated

app/uk/gov/hmrc/hecapplicantfrontend/services/CompanyDetailsService.scala
priyankadhiman3006: updated

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
priyankadhiman3006: done

app/uk/gov/hmrc/hecapplicantfrontend/models/CompanyHouseDetails.scala
priyankadhiman3006: removed

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
```diff
@@ -16,27 +16,139 @@
...
+
+  private def checkCompanyName(crn: CRN, companyRetrievedData: CompanyRetrievedData)(implicit
+    request: RequestWithSessionData[_]
+  ): EitherT[Future, Error, Call] = {
+    val sessionCrn = request.sessionData.userAnswers.fold(_.crn, _.crn)
+    (crn, sessionCrn, companyRetrievedData.companyName) match {
+      //check if the submitted crn is equal to the crn in session and company name is present
+      //then no need to call the companyDetailsService
+      //else fetch company name using the service
+      case (crn, Some(crnSession), Some(_)) if crn.value === crnSession.value =>
```
AndyHWChung, line 104: wouldn't 
```scala
   case (crn, Some(crnSession), None) if(crn.value === crnSession.value =>
```
indicate that we've made the API call before and found no company name? I think we may be able to ignore the `companyName` in session in this pattern match and just base it on the CRN that was just submitted and the one in session

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -280,13 +292,19 @@ object JourneyServiceImpl {
...
             saIncomeDeclared,
-            entityType
+            entityType,
+            _
           ) =>
-        val licenceTypeCheck      = checkEntityTypePresentIfRequired(licenceType, entityType)
-        val saIncomeDeclaredCheck = checkSAIncomeDeclared(taxSituation, saIncomeDeclared, retrievedUserData)
-        licenceTypeCheck && saIncomeDeclaredCheck
+        val licenceTypeCheck = checkEntityTypePresentIfRequired(licenceType, entityType)
+        taxSituation match {
```
AndyHWChung, line 301: perhaps it's be easier to write the function something like:
```scala
val isIndividual = retrieveduserData match {
  case _: IndividualRetrievedData => true 
   case _; CompanyretrievedData => false
}

incompleteUserAnswers match {
   case IncompleteUserAnswers(
            Some(licenceType),
            Some(_),
            Some(_),
            Some(taxSituation),
            saIncomeDeclared,
            entityType
          ) if isIndividual =>
         // ...what was there before...

   case IncompleteUserAnswers(
            Some(licenceType),
            Some(_),
            Some(_),
            _,
            _,
            entityType,
            Some(crn)
          ) if !isIndividual =>
       // ... check entity type as for individuals for now...

    case false =>
} 
```
That way I think you can do `taxSituation` in the pattern match in the uplift method instead of `Some(taxSituation)`

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
priyankadhiman3006: I compared the company name for the case ( which may not occur) where crn matches but company name don't . yes this will never occur  and we can do with just crn  match

priyankadhiman3006: updated

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
priyankadhiman3006: updated

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -152,23 +153,26 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
+          //will add a case for when company when it reaches the complete answers page
+          //if it's  added now, the code thinks the company's answers are all complete and take it to check your answers page
           case incomplete @ IncompleteUserAnswers(
                 Some(licenceType),
                 Some(licenceTimeTrading),
                 Some(licenceValidityPeriod),
                 Some(taxSituation),
                 saIncomeDeclared,
-                entityType
+                entityType,
```
AndyHWChung, line 164: with the update on `allAnswersComplete` I think we can now have `taxSituation` without the `Some` just above here now to make this work for companies now

priyankadhiman3006: if we do it now, then after company's crn page, it think it's all answers are complete and take it to check your answer page. In reality , there will me few more pages after crn and we should change this for company only after all the pages are worked on, what are your thoughts? may be i miss something? There is a comment as well on the top of this case

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
```diff
@@ -16,27 +16,139 @@
...
+      case _                                          =>
+        logger.error("Individual should not see this page")
+        Future.successful(InternalServerError)
+    }
+
+  private def checkCompanyName(crn: CRN, companyRetrievedData: CompanyRetrievedData)(implicit
+    request: RequestWithSessionData[_]
+  ): EitherT[Future, Error, Call] = {
+    val sessionCrn = request.sessionData.userAnswers.fold(_.crn, _.crn)
+    (crn, sessionCrn) match {
```
meghana-viswanath, line 100: don't think there is a need to match on `crn`, just a match on the `sessionCrn` should be enough

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
```diff
@@ -16,27 +16,139 @@
...
+      //check if the submitted crn is equal to the crn in session
+      //then no need to call the companyDetailsService, pick company name from the sesison
+      //else fetch company name using the service
+      case (crn, Some(crnSession)) if crn.value === crnSession.value =>
```
meghana-viswanath, line 104: do we also need to ensure here that along with the crn, the company name is present in the session?

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -232,6 +236,14 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
         }
     }
 
+  private def companyRegistrationNumberRoute(session: HECSession) =
+    session.retrievedUserData match {
+      case IndividualRetrievedData(_, _, _, _, _, _, _) =>
```
meghana-viswanath, line 241:
```suggestion
      case _: IndividualRetrievedData =>
```

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
AndyHWChung: ah ok - I think I finally got you now :smile: When all the company questions are there I think we can still have the one case here since `allAnswersComplete` should cover both individuals and companies. We may forget about this `Some` business here - it's the only thing which is blocking the company journey from completing atm I think and it's not super obvious. One option to make it clearer is to write this case without the `Some` as it would be for the future and then in ` allAnswersComplete` remove the company case so that it always goes to the default `false` case. Then when we get to the part that needs the completing of the answers we can just stick in the separate case there which we need without needing to touch this `uplift` method

priyankadhiman3006: yeah, that was i was thinking earlier and didn't made the company case in allAnswersComplete :). yeah fine with me, i can change  it to as suggested :)

conf/messages
priyankadhiman3006: @nubz Will it be okay for you to resolve this conversation, we can update the code later on when David confirms the error messages. What do you think?

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
priyankadhiman3006: no, because if company name is not present, it means company name was not found in earlier session  and no need to call service again to get Not Found status

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
priyankadhiman3006: we have to match if the crn entered by user through crn formm is same as the crn in the session. for example- user navigate back, and click on continue without changing the crn, then it will check if the crn through form is equal to crn in session or not

meghana-viswanath: yes, definitely, I was only talking about the match, it can be -
```
sessionCrn match {
  case Some(crnSession) if crn.value === crnSession.value => ...
  case _ => ...
}
```

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
meghana-viswanath: ah ok 👍

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
priyankadhiman3006: almost same thing, done differently:) this will reach only if crnSession is Some() :)

AndyHWChung (approved):

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
priyankadhiman3006: [https://github.com/hmrc/hec-applicant-frontend/pull/38/commits/575c0cf3061f0d4afc72e82accfa1edf30a5be8a](updated)

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
priyankadhiman3006: updated

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
meghana-viswanath: seems unnecessary since the `crn` is retained as is in the case match and has the same variable name

conf/messages
nubz: ok by me!

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
priyankadhiman3006: done

meghana-viswanath (approved):

PR #39, opened 2021-09-27
HEC-1107 Start page enhancements

meghana-viswanath:
Details in tickets [HEC-1107](https://jira.tools.tax.service.gov.uk/browse/HEC-1107) & [HEC-1113](https://jira.tools.tax.service.gov.uk/browse/HEC-1113)

- changes to call HEC api and interpret results
- add unexpiredTaxChecks to RetrievedApplicantData
- added route for tax check codes list page
- updates to StartController

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/StartController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/StartController.scala
@@ -16,25 +16,25 @@
 
 package uk.gov.hmrc.hecapplicantfrontend.controllers
 
-import cats.data.{EitherT, Validated}
 import cats.data.Validated.Valid
-import cats.syntax.option._
-import cats.syntax.eq._
-import cats.syntax.traverse._
+import cats.data.{EitherT, Validated}
 import cats.instances.future._
 import cats.instances.option._
 import cats.instances.string._
+import cats.syntax.eq._
+import cats.syntax.option._
+import cats.syntax.traverse._
 import com.google.inject.{Inject, Singleton}
 import play.api.mvc.{Action, AnyContent, MessagesControllerComponents}
-import uk.gov.hmrc.auth.core.{AffinityGroup, ConfidenceLevel, Enrolments}
 import uk.gov.hmrc.auth.core.retrieve.Credentials
+import uk.gov.hmrc.auth.core.{AffinityGroup, ConfidenceLevel, Enrolments}
 import uk.gov.hmrc.hecapplicantfrontend.config.{AppConfig, EnrolmentConfig}
 import uk.gov.hmrc.hecapplicantfrontend.controllers.actions.AuthWithRetrievalsAction
-import uk.gov.hmrc.hecapplicantfrontend.models.{EmailAddress, Error, HECSession, RetrievedApplicantData, RetrievedGGData, UserAnswers}
 import uk.gov.hmrc.hecapplicantfrontend.models.RetrievedApplicantData.{CompanyRetrievedData, IndividualRetrievedData}
 import uk.gov.hmrc.hecapplicantfrontend.models.ids.{CTUTR, GGCredId, NINO, SAUTR}
+import uk.gov.hmrc.hecapplicantfrontend.models.{CitizenDetails, EmailAddress, Error, HECSession, RetrievedApplicantData, RetrievedGGData, TaxCheckListItem, UserAnswers}
 import uk.gov.hmrc.hecapplicantfrontend.repos.SessionStore
-import uk.gov.hmrc.hecapplicantfrontend.services.{CitizenDetailsService, JourneyService}
+import uk.gov.hmrc.hecapplicantfrontend.services.{CitizenDetailsService, JourneyService, TaxCheckService}
 import uk.gov.hmrc.hecapplicantfrontend.util.Logging
 import uk.gov.hmrc.hecapplicantfrontend.util.Logging.LoggerOps
 import uk.gov.hmrc.http.HeaderCarrier
@@ -46,6 +46,7 @@ import scala.concurrent.{ExecutionContext, Future}
 class StartController @Inject() (
   appConfig: AppConfig,
   citizenDetailsService: CitizenDetailsService,
+  taxCheckService: TaxCheckService,
   journeyService: JourneyService,
   sessionStore: SessionStore,
   authWithRetrievalsAction: AuthWithRetrievalsAction,
@@ -126,7 +127,7 @@ class StartController @Inject() (
               )
 
             case _ =>
-              EitherT.fromEither(handleOrganisation(maybeEmail, enrolments, ggCredId))
+              handleOrganisation(maybeEmail, enrolments, ggCredId)
           }
 
         case Some(AffinityGroup.Agent) =>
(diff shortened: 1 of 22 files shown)


AndyHWChung (approved):

priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/controllers/StartController.scala
```diff
@@ -138,6 +139,17 @@ class StartController @Inject() (
     }
   }
 
+  private def fetchTaxChecks()(implicit hc: HeaderCarrier) =
```
AndyHWChung, line 142: if the fetching here fails I don't think we want to proceed - we should treat it as an error rather than saying that there were no tax checks. You should be able to slot the call to `getUnexpiredTaxCheckCodes` inside a for-comprehension around where it's being called I think without having to `fold` on it first - with `EitherT` the idea is build up the `EitherT` and `fold` on the end when you want to handle the overall result.

meghana-viswanath: my thinking was that since the user was on the site to get a new tax check code, the failure of the unexpired tax check codes call shouldn't necessarily be a blocker, that was the reason I didn't build up using an `EitherT`
If that's not what we want, I can definitely change

nubz: worth noting the user _might_ not be on the site to get a new tax check code - they may be on the site to view existing tax check codes... but it's a good point to make about blocking users from getting a new one and perhaps worth following up from a business perspective

AndyHWChung: it's probably safer to build it as an error for now and if we need to change it in response to user feedback later on we can do so then. An error here would most likely be either the BE service is down or mongo is down - if either of things is happening the rest of the journey might not work either.

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxChecksListController.scala
```diff
@@ -0,0 +1,47 @@
...
+    with I18nSupport
+    with Logging {
+
+  /**
+    * Fetches unexpired tax check codes for applicant
+    */
+  val unexpiredTaxChecks: Action[AnyContent] = authAction.andThen(sessionDataAction) { implicit request =>
+    request.sessionData.retrievedUserData.unexpiredTaxChecks match {
+      case Nil       =>
+        logger.warn("No tax chek codes found")
```
priyankadhiman3006, line 41: typo in check

app/uk/gov/hmrc/hecapplicantfrontend/controllers/StartController.scala
meghana-viswanath: @AndyHWChung @nubz just had a slack chat with David, he was of the same idea that it would affect the journey going forward, here's the chat, see what you think

10:15
Meghana Viswanath Hi David, not sure if you have access to github, there was a conversation about what to do when the call to fetch existing unexpired tax check codes fails - should we log the error and continue so user can proceed to generate a new tax check code or do we navigate to the technical error page (or something else)? The ticket is this HEC-1107
10:20
David Murdoch Hi Meghana. This is a technical error, it means something is not working properly with the system.  If you can't pull information back from the DB, you won't be able to get a tax check anyway, you won't be able to have your state saved as you progress through the service.
In some ways its tempting to put up a page saying 'we cant get your existing codes, would you like to generate a new one', but I cant think of what would go wrong with pulling the existing codes back that wouldn't affect generating a new one.
10:21
So even if we allow them to go on to the 'new tax code' process, they will most likely hit a technical error sooner or later anyway.
10:22
Meghana Viswanath ok, yeah, that makes sense, thanks :thumbsup:
10:24
David Murdoch If you think there is a scenario where we cant pull back the existing codes, but could carry on with generating a new code then happy to reconsider. I just cant see an issue with pulling back the codes that wouldn't mean the wider service was also compromised.
10:27
Meghana Viswanath there is always a chance that one call fails but the rest of the system is fine - it could be due to a bug, a temporary system glitch, but more likely than not it's probably safe to assume that there is a system wide issue since we are dealing with the local database in this case, I'll also check with Andy and see what he thinks
New
10:29
David Murdoch Yes, agree, it could just be a bug in that call, making it a localized problem. I would expect a short term database connectivity issue more likely over time, but yes, see what Andy says.

AndyHWChung: that kind of sounds like the direction I was going in I think - an error here is indicative of things not working upstream which may mean that the rest of the journey may not work. The calls to our BE don't happen for a while in the journey - the next would either be API calls for the SA/CT stuff or saving the tax check code - so if the BE is down, if we error out here we can give faster feedback of an error and prevent them from wasting time by going further with the journey. The generic error page says to try again later as well, so if it is a glitch they'd know they'll be able to try again later (or immediately after you see the message if you're a user like me :laughing: )

meghana-viswanath: thanks Andy! updated in commit - https://github.com/hmrc/hec-applicant-frontend/pull/39/commits/72683020ecc501c566c3ebc296bf8303206a609e

app/uk/gov/hmrc/hecapplicantfrontend/controllers/TaxChecksListController.scala
meghana-viswanath: fixed, thanks!

app/uk/gov/hmrc/hecapplicantfrontend/controllers/StartController.scala
AndyHWChung: :+1:  thanks - I think there's a similar change to make in `handleOrganisation` - then you wouldn't need this `fetchTaxChecks` method here

meghana-viswanath: sorry missed, updated here and added unit tests too - https://github.com/hmrc/hec-applicant-frontend/pull/39/commits/f0855fdd12bd030cff746140ffc94271f8ab35d5

priyankadhiman3006: I think ,this chat with david and all comments  should be placed under the ticket comments so that QA people are also aware and is not lost in the PR :)

meghana-viswanath: good shout, I'll add a comment to the ticket now 👍

test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
```diff
@@ -52,29 +52,61 @@ class JourneyServiceSpec extends AnyWordSpec with Matchers with MockFactory with
...
-          val session = HECSession(companyRetrievedData, UserAnswers.empty, None)
-          journeyService.firstPage(session) shouldBe routes.LicenceDetailsController.licenceType()
+          "an individual" in {
+            val individualData = individualRetrievedData.copy(unexpiredTaxChecks = taxChecks)
+            val session        = HECSession(individualData, UserAnswers.empty, None)
+            journeyService.firstPage(session) shouldBe routes.ConfirmIndividualDetailsController
+              .confirmIndividualDetails()
+          }
+
+          "a company" in {
```
priyankadhiman3006, line 86: Shouldn't we have similar test for previous pages as well?

meghana-viswanath: I think that comes in the scope of the [HEC-1099](https://jira.tools.tax.service.gov.uk/browse/HEC-1099) since the scenarios are described there. I was a bit confused about the scope of this ticket too.

priyankadhiman3006: yeah, i thought this to be the ticket which will display the page 🤦‍♀️ .nvm

PR #40, opened 2021-09-28
HEC-1046: update CT status response structure

AndyHWChung:


app/uk/gov/hmrc/hecapplicantfrontend/models/CTStatus.scala
```diff
@@ -29,8 +29,6 @@ object CTStatus {
 
   case object NoticeToFileIssued extends CTStatus
 
-  case object NoAccountingPeriodFound extends CTStatus
```
meghana-viswanath, line 32: does it make sense now to have a single sealed trait for both `SAStatus` and `CTStatus`?

AndyHWChung: it would save some code at this moment in time, but having it separate does also allow for more freedom in the future if we need to do any updates say if one spec adds another status but the other one doesn't. I don't think I'm mega-fussed either way - I guess if we leave it as it is we save some time changing it now and maybe some time in the future if updates do happen.

meghana-viswanath (approved):

priyankadhiman3006 (approved):

PR #42, opened 2021-09-30
HEC-1057 libraries upgraded

priyankadhiman3006:
library dependency changes


project/AppDependencies.scala
```diff
@@ -4,25 +4,25 @@ import sbt._
 object AppDependencies {
 
   val compile = Seq(
-    "uk.gov.hmrc"                %% "bootstrap-frontend-play-28" % "5.3.0",
-    "uk.gov.hmrc"                %% "play-frontend-hmrc"         % "0.89.0-play-28",
+    "uk.gov.hmrc"                %% "bootstrap-frontend-play-28" % "5.14.0",
+    "uk.gov.hmrc"                %% "play-frontend-hmrc"         % "1.14.0-play-28",
```
AndyHWChung, line 8: @nubz  did you check to see if there are any breaking changes that we need to consider for this upgrade?

nubz: Yes - only 2 actions required by all of the changes since 0.89 and one is done here and the other is not applicable - I have pulled and run this branch with new libs and there are no errors

priyankadhiman3006: thanks Ali for confirmation

AndyHWChung: :+1:  thanks

AndyHWChung (approved):

meghana-viswanath (approved):

PR #48, opened 2021-10-07
HEC-1189: correct content on wrong GG account page

AndyHWChung:


AndyHWChung:
I deleted `TmpSpec` while I was here - don't need it anymore since we're actually talking to those API's in the service now

PR #51, opened 2021-10-07
HEC-1887: separate login details and retrieved journey data

AndyHWChung:


priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -226,25 +226,27 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
         sys.error("Could not find tax situation for tax situation route")
       case Some(taxSituation) =>
         if (saTaxSituations.contains(taxSituation)) {
-          session.retrievedUserData match {
-            case IndividualRetrievedData(
-                  IndividualLoginData(_, _, Some(_), _, _, _),
-                  IndividualJourneyData(Some(saStatus)),
-                  _
+          (session.loginData, session.retrievedJourneyData) match {
```
meghana-viswanath, line 229: I didn't anticipate this, it's a bit annoying that we have also match on the login data to ensure that we are working with the right applicant type. Would it be nicer to also have make `RetrievedJourneyData` a sealed trait with case classes for individual/company extending it? Or even take it one level up to `HECSession`, so we have something like the below?
```
sealed trait HECSession {
  val userAnswers: UserAnswers
  val completedTaxCheck: Option[HECTaxCheck]
}

final case class IndividualHECSession(
 loginData: IndividualLoginData,
 retrievedJourneyData: IndividualRetrievedJourneyData,
 userAnswers: UserAnswers,
 completedTaxCheck: Option[HECTaxCheck]
) extends HECSession

final case class CompanyHECSession(
 loginData: CompanyLoginData,
 retrievedJourneyData: CompanyRetrievedJourneyData,
 userAnswers: UserAnswers,
 completedTaxCheck: Option[HECTaxCheck]
) extends HECSession
```

We can have a few helpers of the kind defined in [ConfirmIndividualDetailsController.](https://github.com/hmrc/hec-applicant-frontend/blob/master/app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmIndividualDetailsController.scala#L77) that can be used to ensure we have the right applicant session data. The advantage is that we can always be sure that if we've got a company session, we only have company related data in there.  This block of code might then look something like 
```
withIndividualRetrievedData(request.sessionData) { individualSession =>
  (individualSession.loginData.sautr, individualSession.retrievedJourneyData.saStatus) match {
    case (None, _) => routes.SAController.sautrNotFound()
    case (Some(_), None) => sys.error("Found SA UTR for tax situation route but no SA status response")
    case (Some(utr), Some(saStatus)) =>
      saStatus.status match {
        case SAStatus.ReturnFound => routes.SAController.saIncomeStatement()
        case SAStatus.NoticeToFileIssued => routes.CheckYourAnswersController.checkYourAnswers()
        case SAStatus.NoReturnFound => routes.SAController.noReturnFound()
      }
  }
}
```

AndyHWChung: yeah, I'll do that refactor - I might give the `UserAnswers` refactor a go afterwards and stop if it gets a bit too much

AndyHWChung: updated in [398f2fc](https://github.com/hmrc/hec-applicant-frontend/pull/51/commits/398f2fc656aa0858fb8416653fa151d438eb9e48) 

I think it'd be better to split the user answers into company/individual versions later on just to be able to get on with the actual work. I included the new `taxCheckStartDateTime` field while I was there which will eventually be populated with the start dateTime - it's just sitting there as a `None` for now.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -221,79 +226,81 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
-          case (_, RetrievedJourneyData(_, _, None, _)) =>
-            routes.CompanyDetailsController.ctutrNotMatched()
+  }
 
-          case (CompanyLoginData(_, None, _, _), _) =>
-            routes.CompanyDetailsController.enterCtutr()
+  private def confirmCompanyDetailsRoute(session: HECSession) =
+    session match {
+      case _: IndividualHECSession =>
+        sys.error("This should never happen, individual data shouldn't be present in company journey")
```
meghana-viswanath, line 274: might be useful to have `with/ensureIndividualHECSession` and `with/ensureCompanyHECSession` functions defined somewhere? seems like we'll need to do that match quite a lot

app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmIndividualDetailsController.scala
```diff
@@ -69,17 +69,17 @@ class ConfirmIndividualDetailsController @Inject() (
...
     authAction.andThen(sessionDataAction).async { implicit request =>
-      withIndividualLoginData(request.sessionData) { _ =>
+      withIndividualSesion(request.sessionData) { _ =>
         val back = journeyService.previous(routes.ConfirmIndividualDetailsController.confirmIndividualDetailsExit())
         Ok(confirmIndividualDetailsExitPage(back))
       }
     }
 
-  private def withIndividualLoginData(
+  private def withIndividualSesion(
```
meghana-viswanath, line 78: typo - `withIndividualSession `

app/uk/gov/hmrc/hecapplicantfrontend/models/LoginData.scala
```diff
@@ -16,7 +16,7 @@
 
 package uk.gov.hmrc.hecapplicantfrontend.models
 
-import play.api.libs.json.{JsObject, JsResult, JsValue, Json, OFormat}
+import play.api.libs.json.{Json, OFormat}
 import uk.gov.hmrc.hecapplicantfrontend.models.ids.{CTUTR, GGCredId, NINO, SAUTR}
 
 sealed trait LoginData extends Product with Serializable {
```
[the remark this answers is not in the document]
meghana-viswanath: also, should `unexpiredTaxChecks` be part of login data?

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: yeah, makes sense to pick that at a later point when there's a lull

app/uk/gov/hmrc/hecapplicantfrontend/models/LoginData.scala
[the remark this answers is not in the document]
AndyHWChung: yeah, I was thinking the sealed traits are a bit redundant now but it seemed nice to be able to at relate them together just so they're not completely separate things. `unexpiredTaxChecks` does get populated with the user logs in so I thought there was a connection there. What do you think?

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
AndyHWChung: yeah, I can explore that

app/uk/gov/hmrc/hecapplicantfrontend/models/LoginData.scala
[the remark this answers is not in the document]
AndyHWChung: I had a small go to see what `unexpiredTaxChecks` codes would like like outside - I got to the first hurdle quite quickly. When we create a new session using the `newSession` method we would have to pass in `unexpiredTaxChecks` at the same time but separately cos that's the point where we should have found out what the list should be. That slightly hints that it might belong in `LoginDetails`

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: yeah, naming things is pretty hard, maybe `forIndividual/Company` so at the call site it'll be       
`session.forIndividual { individualSession => ... }`
but I'm not fussed, `mapAs` sounds good too.

app/uk/gov/hmrc/hecapplicantfrontend/models/LoginData.scala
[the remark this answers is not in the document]
AndyHWChung: do you still think the sealed trait thing and the `unexpiredTaxChecks` are still worth doing?

PR #55, opened 2021-10-12
HEC-1048 CT income statement page

meghana-viswanath:
Details in ticket -> https://jira.tools.tax.service.gov.uk/browse/HEC-1048

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
@@ -339,10 +339,9 @@ class CompanyDetailsController @Inject() (
   private def ensureUserAnswersHasCRN(
     session: HECSession
   )(f: CRN => Future[Result]): Future[Result] =
-    session.userAnswers match {
-      case UserAnswers.IncompleteUserAnswers(_, _, _, _, _, _, Some(crn), _, _, _) => f(crn)
-      case UserAnswers.CompleteUserAnswers(_, _, _, _, _, _, Some(crn), _, _, _)   => f(crn)
-      case _                                                                       =>
+    session.userAnswers.fold(_.crn, _.crn) match {
+      case Some(crn) => f(crn)
+      case None      =>
         logger.warn("CRN is not populated in user answers")
         InternalServerError
     }
--- a/app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
@@ -362,14 +362,11 @@ object JourneyServiceImpl {
     val ctStatus = companySession.retrievedJourneyData.ctStatus
       .flatMap(_.latestAccountingPeriod.map(_.ctStatus))
 
-    chargeableForCT match {
-      case YesNoAnswer.Yes =>
-        ctStatus match {
-          case Some(CTStatus.NoticeToFileIssued) => true
-          case Some(_)                           => ctIncomeDeclared.nonEmpty
-          case None                              => false
-        }
-      case YesNoAnswer.No  => true
+    (ctStatus, chargeableForCT) match {
+      case (Some(_), YesNoAnswer.No)                            => true
+      case (Some(CTStatus.NoticeToFileIssued), YesNoAnswer.Yes) => true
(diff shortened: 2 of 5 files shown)


app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -282,9 +340,9 @@ class CompanyDetailsController @Inject() (
     session: HECSession
   )(f: CRN => Future[Result]): Future[Result] =
     session.userAnswers match {
-      case UserAnswers.IncompleteUserAnswers(_, _, _, _, _, _, Some(crn), _, _) => f(crn)
-      case UserAnswers.CompleteUserAnswers(_, _, _, _, _, _, Some(crn), _, _)   => f(crn)
-      case _                                                                    =>
+      case UserAnswers.IncompleteUserAnswers(_, _, _, _, _, _, Some(crn), _, _, _) => f(crn)
```
AndyHWChung, line 343: we could avoid having to update this whenever we add new things be doing something like:
```scala
session.userAnswers.fold(_.crn, _.crn).map(f).getOrElse{
  logger.warn(...)
  InternalServerError
}

File: app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala [hunk 2]
```diff
@@ -347,12 +350,21 @@ object JourneyServiceImpl {
...
-    val ctStatusIsNoticeToFileIssued = companySession.retrievedJourneyData.ctStatus
-      .exists(_.latestAccountingPeriod.exists(_.ctStatus === CTStatus.NoticeToFileIssued))
-
-    (chargeableForCT === YesNoAnswer.No) || (chargeableForCT === YesNoAnswer.Yes && ctStatusIsNoticeToFileIssued)
+    val ctStatus = companySession.retrievedJourneyData.ctStatus
+      .flatMap(_.latestAccountingPeriod.map(_.ctStatus))
+
+    chargeableForCT match {
+      case YesNoAnswer.Yes =>
+        ctStatus match {
```
AndyHWChung, line 361: it feels like the `chargeableForCT` and `ctStatus` pattern match is the wrong way around - `ctStatus` controls whether or not you see the chargeable for CT question. It'd be easier to trace and write the business logic at least if the order of the pattern match reflected the order of the questions.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -396,13 +409,14 @@ object JourneyServiceImpl {
                 Some(_),
                 _,
                 _,
-                entityType,
+                Some(entityType),
```
AndyHWChung, line 412: it'd be safer to leave it as not a `Some` - currently, our licence types require that we always ask the entity type question for companies since our licence types are for individuals only or can be for both entity types. But in the future, if any licence types are added which are company only, then the entity type question won't get asked and it'd be `None` here, but `checkEntityTypePresentIfRequired` should pick up whether or not the entity type has been given or not when expected

conf/app.routes
```diff
@@ -53,6 +53,7 @@ GET        /no-account-period              uk.gov.hmrc.hecapplicantfrontend.cont
 GET        /chargeable-for-ct              uk.gov.hmrc.hecapplicantfrontend.controllers.CompanyDetailsController.chargeableForCorporationTax
 POST       /chargeable-for-ct              uk.gov.hmrc.hecapplicantfrontend.controllers.CompanyDetailsController.chargeableForCorporationTaxSubmit
 GET        /ct-income-statement            uk.gov.hmrc.hecapplicantfrontend.controllers.CompanyDetailsController.ctIncomeStatement
+POST       /ct-income-statement-submit     uk.gov.hmrc.hecapplicantfrontend.controllers.CompanyDetailsController.ctIncomeStatementSubmit
```
AndyHWChung, line 56: in general our submit endpoints have the same url as the corresponding `GET` endpoint - it's not the end of the world with a `-submit` at the end but it'd be neater to have it consistent with the other ones

priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
meghana-viswanath: yeah, I guess I agree, I've refactored the code and also updated the tests so hopefully this scenarios when company check is satisfied is clearer

PR #56, opened 2021-10-13
Fixture functions for unit tests

meghana-viswanath:
I've found that since we avoid using defaults (and correctly so), we end up having to update a majority of the specs everytime we add a new field to the models. To avoid this and reduce the number of files touched, I'm proposing this solution where we create fixture functions with sensible defaults that we can override at the call site. When a new field is added/removed we only have to update the fixture function.

We can add more functions if we are happy with this approach.


meghana-viswanath:
> more rebase for me :'(

sorry! should just be one file though

PR #62, opened 2021-10-21
Hec 1151 cya company

priyankadhiman3006:
ticket -> https://jira.tools.tax.service.gov.uk/browse/HEC-1151

### DIFF
--- a/test/uk/gov/hmrc/hecapplicantfrontend/controllers/CheckYourAnswersControllerSpec.scala
+++ b/test/uk/gov/hmrc/hecapplicantfrontend/controllers/CheckYourAnswersControllerSpec.scala
@@ -182,40 +182,32 @@ class CheckYourAnswersControllerSpec
           )
         }
 
-        "applicant is a Company" in {
-          val answers = Fixtures.completeUserAnswers(
+        "applicant is a Company and " when {
+
+          def createCompleteAnswers(
+            chargeableForCTOpt: Option[YesNoAnswer],
+            ctIncomeDeclaredOpt: Option[YesNoAnswer],
+            recentlyStartedTrading: Option[YesNoAnswer]
+          ) = Fixtures.completeUserAnswers(
             LicenceType.ScrapMetalMobileCollector,
             LicenceTimeTrading.ZeroToTwoYears,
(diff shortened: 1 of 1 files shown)


app/uk/gov/hmrc/hecapplicantfrontend/views/CheckYourAnswers.scala.html
```diff
@@ -28,7 +28,7 @@
 
 @this(layout: Layout, govukSummaryList : GovukSummaryList, govukButton: GovukButton, formWithCSRF: FormWithCSRF)
 
-@(back: Call, completeAnswers: CompleteUserAnswers)(implicit request: Request[_], messages: Messages)
+@(back: Call, completeAnswers: CompleteUserAnswers, date: Option[String])(implicit request: Request[_], messages: Messages)
```
priyankadhiman3006, line 31: This is under discussion, if need to pass Option date here or refactor data model to include date under chargeableForCT

app/uk/gov/hmrc/hecapplicantfrontend/models/TaxDetails.scala
```diff
@@ -32,7 +32,11 @@ object TaxDetails {
   ) extends TaxDetails
 
   final case class CompanyTaxDetails(
-    ctutr: CTUTR
+    ctutr: CTUTR,
```
AndyHWChung, line 35: after this morning's discussion could we change this to have two fields for the CTUTR in here and the BE? Something like:
```scala
  retrievedCTUTR: CTUTR,
  userSuppliedCTUTR: Option[CTUTR]
```
one for the DES CTUTR in `CompanyRetrievedJourneyData` which should always be there and one for the user supplied value if there is one in `UserAnswers`. That will allow us to have an audit of the two different ones if it becomes important later on. The one we'd send in the file transfer is the DES one.

priyankadhiman3006: Is n't this the DESCTUTR
CTStatusResponse(
  ctutr: CTUTR,
  startDate: LocalDate,
  endDate: LocalDate,
  latestAccountingPeriod: Option[CTAccountingPeriod]
)?

PR #73, opened 2021-11-01
fix IV url

AndyHWChung:


LokeshSrinivasa (approved):

app/uk/gov/hmrc/hecapplicantfrontend/config/AppConfig.scala
```diff
@@ -96,7 +97,7 @@ class AppConfig @Inject() (config: Configuration, contactFrontendConfig: Contact
         s"$selfBaseUrl$successRelativeUrl" -> s"$selfBaseUrl$failureRelativeUrl"
     }
 
-    val redirectToIvUrl: String = s"$ivUrl/mdtp/uplift"
+    val redirectToIvUrl: String = s"$ivUrl$ivLocation/uplift"
```
meghana-viswanath, line 100: not sure about this, but should the `/mdtp` bit still be part of the url?

Screenshot 2021-11-01 at 09 44 12

Basing this off the last line in 👆

AndyHWChung: :thinking: I'll double check...I tested it by changing the url in the browser but not the actual changed code

AndyHWChung: seems to work ok - that last line seems to be a bit wrong

meghana-viswanath (approved):

PR #76, opened 2021-11-02
Hec 1185 technical error

priyankadhiman3006:
ticket details -> https://jira.tools.tax.service.gov.uk/browse/HEC-1185


app/uk/gov/hmrc/hecapplicantfrontend/controllers/AccessDeniedController.scala
```diff
@@ -30,8 +30,7 @@ class AccessDeniedController @Inject() (
 ) extends FrontendController(mcc)
     with I18nSupport
     with Logging {
-
   val accessDenied: Action[AnyContent] = authAction { _ =>
-    Ok(s"access denied, session")
+    Ok(s"access denied")
```
meghana-viswanath, line 34: minor - don't need string interpolation here

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -491,7 +453,7 @@ class CompanyDetailsController @Inject() (
       case Some(companyName) => f(companyName)
       case None              =>
         logger.warn("Missing company name")
-        InternalServerError
+        sys.error("Missing company name")
```
meghana-viswanath, line 456: we could use `systemError` function here and for all occurrences below

app/uk/gov/hmrc/hecapplicantfrontend/controllers/LicenceDetailsController.scala
```diff
@@ -259,10 +235,29 @@ class LicenceDetailsController @Inject() (
           )
       case None              =>
         logger.error("Couldn't find licence Type")
-        InternalServerError
+        sys.error("")
```
meghana-viswanath, line 238:
```suggestion
        sys.error("Couldn't find licence Type")
```

app/uk/gov/hmrc/hecapplicantfrontend/controllers/StartController.scala
```diff
@@ -82,22 +82,22 @@ class StartController @Inject() (
       {
         case DataError(msg) =>
           logger.warn(s"Issue with data: $msg")
-          InternalServerError
+          sys.error(s"Issue with data: $msg")
 
         case BackendError(e) =>
           logger.warn(s"Backend error", e)
-          InternalServerError
+          sys.error("")
```
meghana-viswanath, line 89:
```suggestion
          sys.error(s"Backend error - ${e.getMessage}")
```

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
priyankadhiman3006: systemError needs Error parameter as well, which we don't have here

meghana-viswanath: ah, I see it now 👍

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -524,8 +486,27 @@ class CompanyDetailsController @Inject() (
       case Some(crn) => f(crn)
       case None      =>
         logger.warn("CRN is not populated in user answers")
-        InternalServerError
+        sys.error("CRN is not populated in user answers")
     }
+
+  private def systemError(errorMsg: String)(e: Error) = {
```
AndyHWChung, line 492: we probably don't need this - a `sys.error` should give us the message in the logs anyway I think

priyankadhiman3006: in that case , Should i  remove logger.warn wherever we have sys.error?

meghana-viswanath (approved):

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
priyankadhiman3006: updated [here](https://github.com/hmrc/hec-applicant-frontend/pull/76/commits/06ebed2ef02cf1f8cd675f936d86f8bc47458112)
missed changes [here](https://github.com/hmrc/hec-applicant-frontend/pull/76/commits/c307af60b13f49ad132ff5301201667c7b3f1d34)

app/uk/gov/hmrc/hecapplicantfrontend/models/Error.scala
```diff
@@ -24,4 +24,9 @@ object Error {
 
   def apply(error: Throwable): Error = Error(Right(error))
 
+  @SuppressWarnings(Array("org.wartremover.warts.Throw"))
+  implicit class ErrorOps(e: Error) {
+    def throws(message: String): Nothing =
+      e.value.fold(info => sys.error(s"$info::$message"), ex => throw new RuntimeException(message, ex))
```
AndyHWChung, line 30: I think it'd make more sense to have it this way around:
```scala
s"$message::$info"
```
since `message` is usually the more general thing which is broken by `info`. Also, small thing but I think `throw` instead of `throws` makes a bit more sense so it sounds more like a command

priyankadhiman3006: throw is a keyword , so getting compile time error to create a def with same name :(

AndyHWChung (approved):

PR #85, opened 2021-11-17
HEC-1426 CTUTR attempts blocking functionality

meghana-viswanath:
Details in ticket -> https://jira.tools.tax.service.gov.uk/browse/HEC-1426

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
@@ -331,7 +331,7 @@ class CompanyDetailsController @Inject() (
                   companySession.copy(
                     userAnswers = updatedAnswers,
                     retrievedJourneyData = updatedRetrievedData,
-                    ctutrAnswerAttempts = 0
+                    crnBlocked = false
                   )
                 )
             } yield next
@@ -362,7 +362,7 @@ class CompanyDetailsController @Inject() (
                     case ctutrAttempts if ctutrAttempts.isBlocked =>
                       updateAndNextJourneyData(
                         routes.CompanyDetailsController.enterCtutr(),
-                        companySession.copy(ctutrAnswerAttempts = ctutrAttempts.attempts)
+                        companySession.copy(crnBlocked = ctutrAttempts.isBlocked)
                       )
                     case ctutrAttempts                            =>
                       val updatedAnswers = companySession.userAnswers
@@ -372,7 +372,7 @@ class CompanyDetailsController @Inject() (
                         .unset(_.chargeableForCT)
                         .copy(ctutr = formWithErrors.value)
                       val updatedSession = companySession
-                        .copy(ctutrAnswerAttempts = ctutrAttempts.attempts, userAnswers = updatedAnswers)
+                        .copy(crnBlocked = ctutrAttempts.isBlocked, userAnswers = updatedAnswers)
                       sessionStore
                         .store(updatedSession)
                         .foldF(
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
@@ -349,14 +349,13 @@ class CompanyDetailsController @Inject() (
             )
           )
 
-          def handleFormWithErrors(formWithErrors: Form[CTUTR], ctutrAttemptsOpt: Option[CtutrAttempts])
-            : Future[Result] = {
+          def handleFormWithErrors(formWithErrors: Form[CTUTR], ctutrAttempts: CtutrAttempts): Future[Result] = {
 
             def displayFormError: Future[Result] = Future.successful(ok(formWithErrors))
 
             def incrementAttemptsAndProceed =
               ctutrAttemptsService
-                .updateAttempts(crn, companySession.loginData.ggCredId, ctutrAttemptsOpt)
+                .updateAttempts(ctutrAttempts)
                 .foldF(
                   _.doThrow("Could not create/update ctutr attempts"),
                   {
@@ -388,20 +387,20 @@ class CompanyDetailsController @Inject() (
           }
 
           ctutrAttemptsService
-            .get(crn, companySession.loginData.ggCredId)
+            .getWithDefault(crn, companySession.loginData.ggCredId)
             .foldF(
               _.doThrow("Error fetching CTUTR attempts"),
-              {
-                case Some(ctutrAttempts) if ctutrAttempts.isBlocked =>
+              ctutrAttempts =>
+                if (ctutrAttempts.isBlocked) {
                   updateAndNextJourneyData(
                     routes.CompanyDetailsController.enterCtutr(),
                     companySession.copy(crnBlocked = true)
                   )
-                case ctutrAttempts                                  =>
+                } else {
                   enterCtutrForm(desCtutr)
                     .bindFromRequest()
                     .fold(handleFormWithErrors(_, ctutrAttempts), handleValidAnswer)
-              }
+                }
             )
         }
       }
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
@@ -327,8 +327,12 @@ class CompanyDetailsController @Inject() (
               .copy(ctutr = Some(ctutr))
             val (start, end)   = CompanyDetailsController.calculateLookBackPeriod(timeProvider.currentDate)
 
+            val ctutrAttemptsDeleteFut = ctutrAttemptsService.delete(crn, companySession.loginData.ggCredId)
+            val ctStatusFut            = taxCheckService.getCTStatus(ctutr.strippedCtutr, start, end)
+
             val result = for {
-              ctStatus            <- taxCheckService.getCTStatus(ctutr.strippedCtutr, start, end)
+              _                   <- ctutrAttemptsDeleteFut
+              ctStatus            <- ctStatusFut
               updatedRetrievedData = companySession.retrievedJourneyData.copy(ctStatus = ctStatus)
               next                <-
                 journeyService.updateAndNext(
@@ -364,7 +368,7 @@ class CompanyDetailsController @Inject() (
             def incrementAttemptsAndDisplayFormError = {
 
               val result = for {
-                ctutrAttempts <- ctutrAttemptsService.fetchAndUpdateFor(crn, companySession.loginData.ggCredId)
+                ctutrAttempts <- ctutrAttemptsService.createOrIncrementAttempts(crn, companySession.loginData.ggCredId)
                 updatedAnswers = companySession.userAnswers
                                    .unset(_.ctutr)
                                    .unset(_.ctIncomeDeclared)
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
@@ -309,15 +309,6 @@ class CompanyDetailsController @Inject() (
     request.sessionData mapAsCompany { companySession =>
       ensureCompanyDataHasDesCtutr(companySession) { desCtutr =>
         ensureUserAnswersHasCRN(companySession) { crn =>
-          def maxCtutrAnswerAttemptsReached(companySession: CompanyHECSession) =
-            companySession.ctutrAnswerAttempts >= appConfig.maxCtutrAnswerAttempts
-
-          def goToNextPage(updatedSession: CompanyHECSession): Future[Result] =
-            updateAndNextJourneyData(
-              routes.CompanyDetailsController.enterCtutr(),
-              updatedSession
-            )
-
           def handleValidAnswer(ctutr: CTUTR) = {
             val updatedAnswers = companySession.userAnswers
               .unset(_.ctutr)
@@ -360,53 +351,44 @@ class CompanyDetailsController @Inject() (
 
           def handleFormWithErrors(formWithErrors: Form[CTUTR]): Future[Result] = {
 
-            val ctutrsNotMatched = formWithErrors.errors.exists(_.message === "error.ctutrsDoNotMatch")
-            val ctutrOpt         = formWithErrors.value
-
             def displayFormError: Future[Result] = Future.successful(ok(formWithErrors))
 
-            def incrementAttemptsAndDisplayFormError = {
-
-              val result = for {
-                ctutrAttempts <- ctutrAttemptsService.createOrIncrementAttempts(crn, companySession.loginData.ggCredId)
-                updatedAnswers = companySession.userAnswers
-                                   .unset(_.ctutr)
-                                   .unset(_.ctIncomeDeclared)
-                                   .unset(_.recentlyStartedTrading)
-                                   .unset(_.chargeableForCT)
-                                   .copy(ctutr = ctutrOpt)
-                updatedSession = companySession
-                                   .copy(ctutrAnswerAttempts = ctutrAttempts.attempts, userAnswers = updatedAnswers)
-                _             <- sessionStore.store(updatedSession)
-              } yield updatedSession
-
-              result.foldF(
-                _.doThrow("Could not update ctutr answer attempts"),
-                updatedSession =>
-                  //If session is updated and successfully stored in DB, then do a check if ctutr attempt has reached the max limit
-                  // if yes, then go to too many ctutr attempt page else display the form error
-                  //In the absence if this check, form error will display and user will be able to do one more attempt than the max limit.
-                  if (maxCtutrAnswerAttemptsReached(updatedSession)) {
-                    goToNextPage(updatedSession)
-                  } else {
-                    displayFormError
+            def incrementAttemptsAndProceed =
+              ctutrAttemptsService
+                .createOrIncrementAttempts(crn, companySession.loginData.ggCredId)
+                .foldF(
+                  _.doThrow("Could not create/update ctutr attempts"),
+                  {
+                    case CtutrAttempts(_, _, attempts, Some(_)) =>
+                      updateAndNextJourneyData(
+                        routes.CompanyDetailsController.enterCtutr(),
+                        companySession.copy(ctutrAnswerAttempts = attempts)
+                      )
+                    case ctutrAttempts                          =>
+                      val updatedAnswers = companySession.userAnswers
+                        .unset(_.ctutr)
+                        .unset(_.ctIncomeDeclared)
+                        .unset(_.recentlyStartedTrading)
+                        .unset(_.chargeableForCT)
+                        .copy(ctutr = formWithErrors.value)
+                      val updatedSession = companySession
+                        .copy(ctutrAnswerAttempts = ctutrAttempts.attempts, userAnswers = updatedAnswers)
+                      sessionStore
+                        .store(updatedSession)
+                        .foldF(
+                          _.doThrow("Could not save session"),
+                          _ => displayFormError
+                        )
                   }
-              )
-            }
+                )
 
-            if (ctutrsNotMatched && maxCtutrAnswerAttemptsReached(companySession)) goToNextPage(companySession)
-            else if (ctutrsNotMatched) incrementAttemptsAndDisplayFormError
-            else displayFormError
-
+            val ctutrsNotMatched = formWithErrors.errors.exists(_.message === "error.ctutrsDoNotMatch")
+            if (ctutrsNotMatched) incrementAttemptsAndProceed else displayFormError
           }
 
-          if (maxCtutrAnswerAttemptsReached(companySession)) goToNextPage(companySession)
-          else {
-            enterCtutrForm(desCtutr)
-              .bindFromRequest()
-              .fold(handleFormWithErrors, handleValidAnswer)
-          }
-
+          enterCtutrForm(desCtutr)
+            .bindFromRequest()
+            .fold(handleFormWithErrors, handleValidAnswer)
         }
       }
     }
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
@@ -33,7 +33,7 @@ import uk.gov.hmrc.hecapplicantfrontend.models._
 import uk.gov.hmrc.hecapplicantfrontend.models.ids.{CRN, CTUTR}
 import uk.gov.hmrc.hecapplicantfrontend.models.views.YesNoOption
 import uk.gov.hmrc.hecapplicantfrontend.repos.SessionStore
-import uk.gov.hmrc.hecapplicantfrontend.services.{JourneyService, TaxCheckService}
+import uk.gov.hmrc.hecapplicantfrontend.services.{CtutrAttemptsService, JourneyService, TaxCheckService}
 import uk.gov.hmrc.hecapplicantfrontend.util.StringUtils.StringOps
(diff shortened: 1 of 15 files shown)


app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -307,74 +308,81 @@ class CompanyDetailsController @Inject() (
...
+          def handleValidAnswer(ctutr: CTUTR) = {
+            val updatedAnswers = companySession.userAnswers
+              .unset(_.ctutr)
+              .unset(_.ctIncomeDeclared)
+              .unset(_.recentlyStartedTrading)
+              .unset(_.chargeableForCT)
+              .copy(ctutr = Some(ctutr))
+            val (start, end)   = CompanyDetailsController.calculateLookBackPeriod(timeProvider.currentDate)
+
+            val ctutrAttemptsDeleteFut = ctutrAttemptsService.delete(crn, companySession.loginData.ggCredId)
```
priyankadhiman3006, line 330: it could have been included directly into the for comprehension below, so as ctStatusFut

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -307,74 +308,81 @@ class CompanyDetailsController @Inject() (
...
-            val updatedAnswers = companySession.userAnswers
-              .unset(_.ctutr)
-              .unset(_.ctIncomeDeclared)
-              .unset(_.recentlyStartedTrading)
-              .unset(_.chargeableForCT)
-              .copy(ctutr = ctutrOpt)
-            val updatedSession = companySession
-              .copy(ctutrAnswerAttempts = companySession.ctutrAnswerAttempts + 1, userAnswers = updatedAnswers)
+              val result = for {
+                ctutrAttempts <- ctutrAttemptsService.createOrIncrementAttempts(crn, companySession.loginData.ggCredId)
```
AndyHWChung, line 371: if the maximum number of attempts has already been reached then I think with the current set up the `lastUpdated` field in `CtutrAttempts` will be updated. We'll eventually need the time where the lock expires, so maybe instead of
```scala
final case class CtutrAttempts(
  crn: CRN,
  ggCredId: GGCredId,
  attempts: Int,
  lastUpdated: ZonedDateTime
)
```
we could have something like
```scala
final case class CtutrAttempts(
  crn: CRN,
  ggCredId: GGCredId,
  attempts: Int,
  lockedUntil: Option[ZonedDateTime]
)
```
Where `lockedUntil` is populated as soon as the maximum number of attempts has been reached and not updated any time afterwards. That means we'd be able to tell if someone has been locked out by looking at the presence of that field instead of looking at the number of attempts - this means it'd be more resilient to changes of the maximum value if we ever change it - people who are locked out under an old value will still be locked out if we increase the value. Also, when we come to display the time they can try again, we don't need to do a calculation based upon the last updated time.

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -307,74 +308,81 @@ class CompanyDetailsController @Inject() (
...
+            val updatedAnswers = companySession.userAnswers
+              .unset(_.ctutr)
+              .unset(_.ctIncomeDeclared)
+              .unset(_.recentlyStartedTrading)
+              .unset(_.chargeableForCT)
+              .copy(ctutr = Some(ctutr))
+            val (start, end)   = CompanyDetailsController.calculateLookBackPeriod(timeProvider.currentDate)
+
+            val ctutrAttemptsDeleteFut = ctutrAttemptsService.delete(crn, companySession.loginData.ggCredId)
+            val ctStatusFut            = taxCheckService.getCTStatus(ctutr.strippedCtutr, start, end)
```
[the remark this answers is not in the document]
meghana-viswanath: My thinking was to kick both the futures off simultaneously and save time, doing them directly in the for-comprehension will mean waiting for the first future before doing the second future. Agree that in the case that one of them fails, the other will be done, but those are exceptional cases and won't happen as often, so leaned towards saving time here

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
meghana-viswanath: my reasoning [here](https://github.com/hmrc/hec-applicant-frontend/pull/85#discussion_r751096891), we can continue the discussion there?

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
meghana-viswanath: yeah, I see your point, we can get hold of `lastUpdated` anyway from the context data, so makes sense to pre-calculate and store 👍

app/uk/gov/hmrc/hecapplicantfrontend/services/CtutrAttemptsService.scala
```diff
@@ -0,0 +1,68 @@
...
+  ctutrAttemptsStore: CtutrAttemptsStore,
+  timeProvider: TimeProvider
+)(implicit
+  ec: ExecutionContext
+) extends CtutrAttemptsService {
+
+  /**
+    * Fetch CTUTR attempts for the CRN-GGCredId combination
+    * If found, increment the number of attempts
+    * If not found, create a new document with 0 attempts
```
priyankadhiman3006, line 48: do we need to create a new doc with 0 attempts? cause as per the discussion during refinement when we are checking the db and  there is no data it means it is not blocking. I may be wrong , but just want to double check

meghana-viswanath: no, you are right, the docstring is incorrect, it creates a new doc and sets the number of attempts to 1, I'll fix the comment 👍

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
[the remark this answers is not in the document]
AndyHWChung: ah ok - I get your logic here. In theory this shouldn't cause a lot of issues cos the call to delete failing is going to be rare, so a wasted call to the CT status API is rare. I think I'm ok leaving this as is then and seeing if it causes any issues in the future (which I don't think will be the case). What do you think @priyankadhiman3006?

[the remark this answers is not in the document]
priyankadhiman3006: yup, this will not cause a issue  , just the call to fetch ctStats will get wasted , even if the delete fails. no issues in keeping the existing code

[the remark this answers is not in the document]
priyankadhiman3006: I was wondering, when should we delete the attempts from the DB, should it be done in the beginning and then session is updated as done here , or should be done at the end , so even if the db delete is failed, we have correct value in the session?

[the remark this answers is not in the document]
meghana-viswanath: not sure I understand, as it is now, if the delete fails, a `RuntimeException` is thrown and the session is left as is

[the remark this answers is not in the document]
priyankadhiman3006: nvm, pls ignore

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
meghana-viswanath: updated https://github.com/hmrc/hec-applicant-frontend/pull/85/commits/5fc7cc1a19ceda4df7804ce3510a700a4b57dcde

app/uk/gov/hmrc/hecapplicantfrontend/services/CtutrAttemptsService.scala
meghana-viswanath: updated https://github.com/hmrc/hec-applicant-frontend/pull/85/commits/5fc7cc1a19ceda4df7804ce3510a700a4b57dcde

app/uk/gov/hmrc/hecapplicantfrontend/services/CtutrAttemptsService.scala
```diff
@@ -0,0 +1,76 @@
...
+    * Store the new/updated data back into the database
+    * @return The updated CTUTR attempts
+    */
+  def createOrIncrementAttempts(crn: CRN, ggCredId: GGCredId): EitherT[Future, models.Error, CtutrAttempts] =
+    for {
+      attemptsOpt    <- ctutrAttemptsStore.get(crn, ggCredId)
+      updatedAttempts = attemptsOpt match {
+                          case Some(ctutrAttempts) =>
+                            val lockedUntilOpt = if (ctutrAttempts.attempts >= appConfig.maxCtutrAnswerAttempts - 1) {
+                              val durationInSeconds = appConfig.maxCtutrStoreExpiry.toSeconds
```
AndyHWChung, line 62: `appConfig.maxCtutrStoreExpiry.toSeconds` could be taken out of this `def` so we don't need to do the calculation to seconds every time

test/uk/gov/hmrc/hecapplicantfrontend/utils/Fixtures.scala
```diff
@@ -205,4 +209,36 @@ object Fixtures {
     ctStatus = ctStatus
   )
 
+  class TestConfig(
```
AndyHWChung, line 212: this seems like quite a big thing to maintain - we could maybe explore using scala mock to create a mock instance of AppConfig and configure it to return certain values of val's like we do in the controller specs

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -307,74 +308,81 @@ class CompanyDetailsController @Inject() (
...
-          companySession.ctutrAnswerAttempts >= appConfig.maxCtutrAnswerAttempts
-
-        def goToNextPage(updatedSession: CompanyHECSession): Future[Result] =
-          updateAndNextJourneyData(
-            routes.CompanyDetailsController.enterCtutr(),
-            updatedSession
-          )
+        ensureUserAnswersHasCRN(companySession) { crn =>
+          def maxCtutrAnswerAttemptsReached(companySession: CompanyHECSession) =
+            companySession.ctutrAnswerAttempts >= appConfig.maxCtutrAnswerAttempts
```
AndyHWChung, line 313: this check doesn't work for different CRN's now - I don't think we need `ctutrAnswerAttempts` in session anymore. Here, we could do a get on the `CtutrAttemptsStore` with the `CRN` and `ggCredId` and check if `lockedUntil` exists to see if the maximum number of attempts has been reached for this particular CRN. If we do that, then this controller doesn't need to know the value of `appConfig.maxCtutrAnswerAttempts` and then we can remove the CT UTR attempts config values from `AppConfig` and just inject `Configuration` into `CtutrAttemptsService`. Then we can leave `AppConfig` as it was before

test/uk/gov/hmrc/hecapplicantfrontend/utils/Fixtures.scala
meghana-viswanath: Yeah, we would have to add a new field/method in 3 places as opposed to one. In controller specs we extend `ControllerSpec` and override the `additionalConfig` like this - 
```
override lazy val additionalConfig: Configuration = Configuration(
    ConfigFactory.parseString(s"ctutr-attempts.maximum-attempts = $maxCtutrAttempts")
  )
```
But here, we want to inject `AppConfig` which is not the same.

The problem is that `scalamock` does not allow mocking `val`s - their workaround suggestion is to use `def`s instead and I didn't want to do that. If we did use `def`s, we can do the below and get rid of 👆 
```
val mockAppConfig = mock[AppConfig]
(mockAppConfig.maxCtutrAnswerAttempts _).expects().returning(3)
(mockAppConfig.maxCtutrStoreExpiry _).expects().returning(3 hours)
```

Gladly welcome any ideas here because I'm not super happy with this either.

AndyHWChung: I think my suggestion in my other comment might help where we may be able to circumvent the need for injecting in `AppConfig` entirely

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
meghana-viswanath: think you are right, I'll give a go, I think we'll still need something in the session to indicate the presence of a lock so the `JourneyService` can determine what to do next

meghana-viswanath:
@AndyHWChung @priyankadhiman3006  I've made the following changes based on comments from yesterday
- replaced `CtutrAttempts.lastUpdated` with `blockedUntil` that gets populated only when the max allowed attempts is reached. Once this field is populated, the value in ctutr attempts field is irrelevant.
- instead of injecting `AppConfig` into `CtutrAttemptsServiceImpl`, I've taken Andy's suggestion and injected `Configuration` instead allowing us to revert to the original simpler state of `AppConfig`
- replaced `CompanyHECSession.ctutrAttempts` with `crnBlocked` boolean field. We need a way to understand when a CRN is blocked within the `JourneyService` so this information needs to be in the session. But a simple boolean indicating whether or not the CRN in the answers is blocked should be sufficient.

Let me know your thoughts.

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -307,99 +308,88 @@ class CompanyDetailsController @Inject() (
...
+            def incrementAttemptsAndProceed =
+              ctutrAttemptsService
+                .createOrIncrementAttempts(crn, companySession.loginData.ggCredId)
+                .foldF(
+                  _.doThrow("Could not create/update ctutr attempts"),
+                  {
+                    case ctutrAttempts if ctutrAttempts.isBlocked =>
+                      updateAndNextJourneyData(
+                        routes.CompanyDetailsController.enterCtutr(),
+                        companySession.copy(crnBlocked = ctutrAttempts.isBlocked)
```
AndyHWChung, line 365: `isBlocked` and the `crnBlocked` needs to know which CRN is being referred to - you can in theory block yourself on one CRN but still have three goes on another CRN. 

That may be that `crnBlocked` becomes something like `blockedCRNs: List[CRN]`. At some point it seems like 
the thing in session needs to be cleared since it'll not be blocked forever. So maybe it needs to be more something like `blockedCRNs: List[(CRN, ZonedDateTime)]` where it some case class instead of a tuple and the time refers to the time the block is lifted. Then you can know for certain when you're in the `JourneyService` that the lock hasn't expired.

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -307,99 +308,88 @@ class CompanyDetailsController @Inject() (
...
+                  }
+                )
+
+            val ctutrsNotMatched = formWithErrors.errors.exists(_.message === "error.ctutrsDoNotMatch")
+            if (ctutrsNotMatched) incrementAttemptsAndProceed else displayFormError
+          }
 
-        if (maxCtutrAnswerAttemptsReached(companySession)) goToNextPage(companySession)
-        else {
           enterCtutrForm(desCtutr)
```
priyankadhiman3006, line 389: There should be a check before form submit, if there is a combination of crn + ggcrcedid in db , the form should not look for handling errors or check for valid answer, it should directly go to too many attempt page

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
meghana-viswanath: `isBlocked` is against `CRN` and `GGCredId` combination, so we should be ok there I think.

The `crnBlocked` field on the session will be against the `CRN` in the user answers. When the answer changes, the `crnBlocked` also changes with it by fetching it from mongo and setting it on the session so it can be used in the `JourneyService`. I'm not certain that we need a list in the session.

AndyHWChung: ah ok, yeah I see with `isBlocked` now.

With `crnBlocked` wouldn't we need some changes in `CRNControlelr` on the submit endpoint then to make it reset so we know for sure that `crnBlocked` is for the CRN in session?

priyankadhiman3006:
I think there is a bug in there, go through company journey, enter ctutr wrong two times,
enter it correct 3rd time. go to CYA page. click on change link against ctutr. it still allow you to enter wrong ctutr three more times.
As discussed earlier, We are planning to remove that change link against ctutr  from CYA page, @AndyHWChung should we consider this as bug or ignore it

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
meghana-viswanath: There's another ticket [HEC-1427](https://jira.tools.tax.service.gov.uk/browse/HEC-1427) which has enhancements to the `CRNController` where we do a check to see if the CRN is blocked and proceed based on that. For now, if CRN answer changes, the `crnBlocked` is reset to `false` and user is allowed to carry on

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
meghana-viswanath: ok, I see, makes sense, I'll fix 👍

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
AndyHWChung: to reset it to `false` wouldn't you need something like `session.copy(crnBlocked = false)` in the `CRNController` submit endpoint [here](https://github.com/hmrc/hec-applicant-frontend/blob/mv-hec-1426-blocking/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala#L96)

meghana-viswanath: yeah, that's right, it doesn't get reset, I got confused, but when we've implemented the HEC-1427, we'll do it then, does that sound ok?

so when the CRN answer changes, in the `CRNController`, we check if there's a block against it and set the `crnBlocked` values based on that

meghana-viswanath: would it make sense to do that ticket along with this work?

AndyHWChung: We could do it in either I guess, it should get done eventually. In 1427, in scenario 1 there's mention of being able to proceed if the CRN is not blocked, but there's no mention of being able to do a number of attempts on the CT UTR - perhaps that should be added in that scenario to make sure that that gets tested

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
meghana-viswanath: updated https://github.com/hmrc/hec-applicant-frontend/pull/85/commits/cef49808c78deff4bf0a52ebe269d495f069d189

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
meghana-viswanath: I'll do a separate PR if that's ok @AndyHWChung

app/uk/gov/hmrc/hecapplicantfrontend/services/CtutrAttemptsService.scala
```diff
@@ -0,0 +1,105 @@
...
+      ctutrAttempts.forall(ca => ca.crn === crn && ca.ggCredId === ggCredId),
+      "crn and ggCredId should match corresponding values in ctutrAttempts if defined"
+    )
+
+    val updatedAttempts = ctutrAttempts match {
+      case None                                           => CtutrAttempts(crn, ggCredId, 1, None)
+      case Some(ctutrAttempts) if ctutrAttempts.isBlocked => ctutrAttempts
+      case Some(ctutrAttempts)                            =>
+        val lockedUntilOpt = if (ctutrAttempts.attempts >= maxCtutrAnswerAttempts - 1) {
+          Some(timeProvider.now.plusSeconds(maxCtutrStoreExpiryInSeconds))
```
AndyHWChung, line 89: if a CRN has already been locked, it seems to make more sense to leave the lock time alone until it expires - with this it seems to extend the lock time each time this method gets called. Perhaps this should be:
```scala
ctutrAttempts.blockedUntil
  .OrElse(Some(timeProvider.now.plusSeconds(maxCtutrStoreExpiryInSeconds)))
```

app/uk/gov/hmrc/hecapplicantfrontend/services/CtutrAttemptsService.scala
```diff
@@ -0,0 +1,105 @@
...
+    * @param ggCredId The GGCredId
+    * @param ctutrAttempts Previously fetched CTUTR attempts
+    * @return
+    */
+  def updateAttempts(
+    crn: CRN,
+    ggCredId: GGCredId,
+    ctutrAttempts: Option[CtutrAttempts]
+  ): EitherT[Future, models.Error, CtutrAttempts] = {
+    require(
```
AndyHWChung, line 79: this is a bit ugly having to do it this way - I wonder if there's a way to make the types encode the infromation better. What do you think about `get` not returning `Option[CtutrAttempts]` but returning just `CtutrAttempts` instead with `attempts` equal to `0` if it doesn't exist? Then we could just pass in `CtutrAttempts` attempts to this function without having to worry about mismatching CRN's and GG cred id's

priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/services/CtutrAttemptsService.scala
meghana-viswanath: I like it!

app/uk/gov/hmrc/hecapplicantfrontend/services/CtutrAttemptsService.scala
meghana-viswanath: if CRN has already been locked, the first case statement is matched where it does nothing and returns as is

AndyHWChung: oh yeah - sorry! Ignore me

app/uk/gov/hmrc/hecapplicantfrontend/services/CtutrAttemptsService.scala
meghana-viswanath: updated https://github.com/hmrc/hec-applicant-frontend/pull/85/commits/9a2cf72f8532dd7b9270caebb01a4f4b053b1845

AndyHWChung: :+1:  looks neater now!

AndyHWChung (approved):

app/uk/gov/hmrc/hecapplicantfrontend/services/CtutrAttemptsService.scala
meghana-viswanath: Awesome! Thanks for all the suggestions @AndyHWChung @priyankadhiman3006 !

PR #86, opened 2021-11-17
Hec 1449 language toggle

priyankadhiman3006:
ticket details -> https://jira.tools.tax.service.gov.uk/browse/HEC-1449


conf/application.conf
```diff
@@ -136,12 +136,12 @@ controllers {
 play.i18n.langCookieHttpOnly: "true"
 
 # Change this value to true to enable Welsh translations to be loaded from messages.cy, and to display the language toggle
-features.welsh-language-support = false
+features.welsh-language-support = true
```
AndyHWChung, line 139: Can we change app-config-prod to set this value to false please? Just so it's already done before we launch

priyankadhiman3006: sure

AndyHWChung (approved):

PR #87, opened 2021-11-19
HEC-1427 Check if CRN is blocked in CRN submit endpoint

meghana-viswanath:
Details in ticket -> https://jira.tools.tax.service.gov.uk/browse/HEC-1427

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
@@ -17,7 +17,6 @@
 package uk.gov.hmrc.hecapplicantfrontend.controllers
 
 import cats.data.EitherT
-import cats.implicits.{catsKernelStdOrderForString, catsSyntaxEq}
 import cats.instances.future._
 import cats.syntax.option._
 import com.google.inject.Inject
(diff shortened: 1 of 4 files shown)


AndyHWChung (approved):

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
```diff
@@ -104,15 +108,25 @@ class CRNController @Inject() (
...
+                               updateAndNext(updatedSession)
+                             } else if (crnsMatch) {
+                               updateAndNext(companySession.copy(crnBlocked = false))
+                             } else {
+                               fetchCompanyNameAndProceed(crn)
+                             }
+          } yield eitherResult
+
           result.fold(
             _.doThrow("Could not update session and proceed"),
```
priyankadhiman3006, line 131: can we do a check at  crnForm()
          .bindFromRequest() if crn  blocked  is already in sessioni, then no need to fetch all details and directly show page blocked?

priyankadhiman3006: ok , ii think we shouldn't because we need to check the crn on which the block is there, all good

priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/controllers/CRNController.scala
meghana-viswanath: yeah, we need to allow the user to enter a new CRN if they wish to, otherwise they would always get redirected to the too-many-ctutr-attempts page once they have a blocked CRN in session and would never be allowed to input a new CRN

PR #95, opened 2021-12-07
Enforce max tax checks limit

meghana-viswanath:
When someone submits an answer on the licence type question in the applicant journey we'd do something like:
- work out how many tax check codes the user (identified by their gg cred id) has for the licence type they've just submitted
- if the number is more than a configured maximum then redirect to some exit page saying they've done too many tax checks for that licence type - it's a dummy page for now

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
@@ -262,7 +262,7 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
           .mapValues(_.length)
           .getOrElse(licenceType, 0)
 
-        if (taxCheckCountForLicenceType > appConfig.maxTaxChecksAllowed) {
+        if (taxCheckCountForLicenceType > appConfig.maxTaxChecksPerLicenceType) {
           routes.LicenceDetailsController.maxTaxChecksExceeded()
         } else {
           routes.LicenceDetailsController.licenceTimeTrading()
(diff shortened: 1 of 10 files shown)


conf/application.conf
```diff
@@ -208,4 +208,5 @@ email-allow-list {
 
 auth-login-stub.base-url = "http://localhost:9949"
 
-
+# TODO verify if one number can be used for all licence types
```
AndyHWChung, line 211: the idea was that the max number was per licence type and that it'd be a single number for all licence types I think - maybe naming the parameter name to something like `max-tax-checks-per-licence-type` would make that super clear

meghana-viswanath: updated https://github.com/hmrc/hec-applicant-frontend/pull/95/commits/3ef8edff7313204e0c68581043d028754498709a

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -247,6 +249,28 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
 
+  private def licenceTypeRoute(session: HECSession): Call = {
+    val licenceTypeOpt = session.userAnswers.fold(
+      _.fold(_.licenceType, _.licenceType.some),
+      _.fold(_.licenceType, _.licenceType.some)
+    )
+
+    licenceTypeOpt match {
+      case Some(licenceType) =>
+        val taxCheckCountForLicenceType = session.unexpiredTaxChecks
```
priyankadhiman3006, line 260: unexpiredTaxChecks is calculated at the start, are we taking the licence type in the user answers into account too?

priyankadhiman3006 (approved):

AndyHWChung (approved):

PR #97, opened 2021-12-09
Reorganise models

meghana-viswanath:
| Before      | After |
| ----------- | ----------- |
|  Screenshot 2021-12-09 at 11 37 43| Screenshot 2021-12-09 at 11 37 24|

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/connectors/HECConnector.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/connectors/HECConnector.scala
@@ -18,8 +18,9 @@ package uk.gov.hmrc.hecapplicantfrontend.connectors
 
 import cats.data.EitherT
 import com.google.inject.{ImplementedBy, Inject, Singleton}
+import uk.gov.hmrc.hecapplicantfrontend.models.hecTaxCheck.HECTaxCheckData
 import uk.gov.hmrc.hecapplicantfrontend.models.ids.{CRN, CTUTR, SAUTR}
-import uk.gov.hmrc.hecapplicantfrontend.models.{Error, HECTaxCheckData, TaxYear}
+import uk.gov.hmrc.hecapplicantfrontend.models.{Error, TaxYear}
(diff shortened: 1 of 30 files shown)


app/uk/gov/hmrc/hecapplicantfrontend/connectors/HECConnector.scala
```diff
@@ -18,8 +18,9 @@ package uk.gov.hmrc.hecapplicantfrontend.connectors
 
 import cats.data.EitherT
 import com.google.inject.{ImplementedBy, Inject, Singleton}
+import uk.gov.hmrc.hecapplicantfrontend.models.hecTaxCheck.HECTaxCheckData
```
AndyHWChung, line 21: this isn't to do with this `import`, but what do you think about maybe having a package called `userDetails` or something that put some structure to some of the remaining stuff at the top level in `models` like:
```
userDetails
- EmailAddress
- EntityType
- LoginData
- RetrievedGGData (?)
- RetrievedJourneyData (?)
- individual
  - CitizenDetails
  - DateOfBirth
  - Name
-company
  - CompanyHouseDetails
  - CompanHouseName
```

AndyHWChung: I'm not overly fussed about it - maybe @priyankadhiman3006 you can have the deciding vote here :smile:

priyankadhiman3006: With  the number of case classes we have , I feel we don't need to create that many sub packages.
it will be a lot of sub packaging 😅.

priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/connectors/HECConnector.scala
AndyHWChung: :+1: let's leave it then

AndyHWChung (approved):

PR #99, opened 2021-12-09
Access denied page

meghana-viswanath:
Details in ticket -> [HEC-1576](https://jira.tools.tax.service.gov.uk/browse/HEC-1576)

I found that the page renders correctly when accessed directly via the endpoint, but when it gets redirected via the `EmailAllowedListFilter` due to user not being in allow list, the style does nor render correctly, it looks something like this -

Screenshot 2021-12-09 at 11 07 26

and I see these error in the console - Screenshot 2021-12-09 at 11 10 37

@nubz's thoughts
> I think the email allow list filter is being applied to all requests except just this page/document (which then goes on to make multiple requests but each of those requests is being redirected to this page
> so when I look at the header for a stylesheet request it is being redirected to /tax-check-for-licence/denied-access in the response
> which returns text/html hence the console errors

Any ideas about why this might be so are welcome.

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/filters/EmailAllowedListFilter.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/filters/EmailAllowedListFilter.scala
@@ -44,9 +44,9 @@ class EmailAllowedListFilter @Inject() (
   val userEmailAllowedList: List[String] = config.underlying.get[List[String]]("email-allow-list.list").value
 
   private def isExcludedEndpoint(rh: RequestHeader): Boolean =
-    rh.uri.contains(routes.AccessDeniedController.accessDenied().url) ||
-      rh.uri.contains("hmrc-frontend") ||
-      rh.uri.contains("assets")
+    rh.path.contains(routes.AccessDeniedController.accessDenied().url) ||
+      rh.path.contains("hmrc-frontend") ||
+      rh.path.contains("assets")
 
   override def apply(f: RequestHeader => Future[Result])(rh: RequestHeader): Future[Result] =
     if (userEmailListEnabled) {
--- a/app/uk/gov/hmrc/hecapplicantfrontend/filters/EmailAllowedListFilter.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/filters/EmailAllowedListFilter.scala
@@ -44,7 +44,9 @@ class EmailAllowedListFilter @Inject() (
   val userEmailAllowedList: List[String] = config.underlying.get[List[String]]("email-allow-list.list").value
 
   private def isExcludedEndpoint(rh: RequestHeader): Boolean =
-    rh.uri.contains(routes.AccessDeniedController.accessDenied().url)
+    rh.uri.contains(routes.AccessDeniedController.accessDenied().url) ||
+      rh.uri.contains("hmrc-frontend") ||
+      rh.uri.contains("assets")
 
   override def apply(f: RequestHeader => Future[Result])(rh: RequestHeader): Future[Result] =
     if (userEmailListEnabled) {
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -202,7 +202,7 @@ ctutr-attempts {
 }
 
 email-allow-list {
-  enabled = false
+  enabled = true
   list = [ "user@test.com" ]
 }
 
--- a/conf/application.conf
+++ b/conf/application.conf
@@ -202,7 +202,7 @@ ctutr-attempts {
 }
 
 email-allow-list {
-  enabled = true
+  enabled = false
   list = [ "user@test.com" ]
 }
 
--- a/conf/messages
+++ b/conf/messages
@@ -317,3 +317,11 @@ tooManyCtutrAttempts.crn=Company Registration Number
 # Max tax checks limit exceeded
 maxTaxChecksLimitExceeded.title = Maximum number of tax checks allowed for chosen licence type has been exceeded
 maxTaxChecksLimitExceeded.p1 = Maximum number of tax checks allowed for chosen licence type has been exceeded.
+
+# Access Denied
+accessDenied.title = You cannot access this page
+accessDenied.p1 = If you have been sent this link and cannot access it, you may be logged in with the wrong Government Gateway account.
+accessDenied.p2 = If you need any assistance accessing this link, please email:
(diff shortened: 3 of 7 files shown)


AndyHWChung:
I think we may need to add a couple more things to `isExcludedEndpoint` here so that the page can retrieve what it needs to:
https://github.com/hmrc/hec-applicant-frontend/blob/main/app/uk/gov/hmrc/hecapplicantfrontend/filters/EmailAllowedListFilter.scala#L47

it might be that you need to allow requests such as these through the filter:
```
->         /hmrc-frontend           hmrcfrontend.Routes
GET        /assets/*file            controllers.Assets.versioned(path = "/public", file: controllers.Assets.Asset)
```
If that's the case then you'd need to have some rules like `if(rh.uri.cotnains("/hmrc-frontend")` or `if(rh.uri.contains("/asset") I think in `isExcludedEndpoint`.

conf/application.conf
```diff
@@ -202,7 +202,7 @@ ctutr-attempts {
 }
 
 email-allow-list {
-  enabled = false
+  enabled = true
```
AndyHWChung, line 205: just so we don't need to worry about it locally it'd be handy to keep this `false` by default

meghana-viswanath:
> I think we may need to add a couple more things to `isExcludedEndpoint` here so that the page can retrieve what it needs to: https://github.com/hmrc/hec-applicant-frontend/blob/main/app/uk/gov/hmrc/hecapplicantfrontend/filters/EmailAllowedListFilter.scala#L47
> 
> it might be that you need to allow requests such as these through the filter:
> 
> ```
> ->         /hmrc-frontend           hmrcfrontend.Routes
> GET        /assets/*file            controllers.Assets.versioned(path = "/public", file: controllers.Assets.Asset)
> ```
> 
> If that's the case then you'd need to have some rules like `if(rh.uri.cotnains("/hmrc-frontend")` or `if(rh.uri.contains("/asset") I think in `isExcludedEndpoint`.

That worked @AndyHWChung, updated https://github.com/hmrc/hec-applicant-frontend/pull/99/commits/2a266efc27f26dc7e4b9ee8a353e20659fe5f8c1

conf/application.conf
meghana-viswanath: updated https://github.com/hmrc/hec-applicant-frontend/pull/99/commits/96bbce215534cb0f53ccd484bd87bc125ebf1eb0

conf/messages
```diff
@@ -317,3 +317,11 @@ tooManyCtutrAttempts.crn=Company Registration Number
 # Max tax checks limit exceeded
 maxTaxChecksLimitExceeded.title = Maximum number of tax checks allowed for chosen licence type has been exceeded
 maxTaxChecksLimitExceeded.p1 = Maximum number of tax checks allowed for chosen licence type has been exceeded.
+
+# Access Denied
+accessDenied.title = You cannot access this page
+accessDenied.p1 = If you have been sent this link and cannot access it, you may be logged in with the wrong Government Gateway account.
+accessDenied.p2 = If you need any assistance accessing this link, please email:
+accessDenied.p3 = <a href="mailto:zaynah.hawa@digital.hmrc.gov.uk?subject=Private Beta link assistance">zaynah.hawa@digital.hmrc.gov.uk</a>
```
priyankadhiman3006, line 325: are we sure we need to add the email id of zaynah and alex  😄

priyankadhiman3006 (approved):

priyankadhiman3006:
Shouldn't we have test case for this page? atleast for asserting the title

conf/messages
meghana-viswanath: I checked with Alice and David, and that's the finalised content. Alice said they were trying to get a generic email to share but it was impossible with google.

app/uk/gov/hmrc/hecapplicantfrontend/filters/EmailAllowedListFilter.scala
```diff
@@ -44,7 +44,9 @@ class EmailAllowedListFilter @Inject() (
   val userEmailAllowedList: List[String] = config.underlying.get[List[String]]("email-allow-list.list").value
 
   private def isExcludedEndpoint(rh: RequestHeader): Boolean =
-    rh.uri.contains(routes.AccessDeniedController.accessDenied().url)
+    rh.uri.contains(routes.AccessDeniedController.accessDenied().url) ||
+      rh.uri.contains("hmrc-frontend") ||
+      rh.uri.contains("assets")
```
nubz, line 49: what about a url like `https://www.development.tax.service.gov.uk/tax-check-for-licence/start?hack=assets` could that bypass the allow list?

meghana-viswanath:
> Shouldn't we have test case for this page? atleast for asserting the title

It's a static page, so didn't think we needed tests, but I can add quickly just to have it in place if the content changes 👍

app/uk/gov/hmrc/hecapplicantfrontend/filters/EmailAllowedListFilter.scala
AndyHWChung: :thinking: does `uri` contain the query parameters if there are any? If we somehow exclude query parameters then we should be safe cos then `assets` and`hmrc-frontend` would need to be part of a url we've configured in our service

nubz: I think you are right - wasn't sure how request header broke down - uri should not contain query params but url would - I tested it and got blocked so it's all good...

AndyHWChung (approved):

app/uk/gov/hmrc/hecapplicantfrontend/filters/EmailAllowedListFilter.scala
meghana-viswanath: I just looked at the source code and it looks like the uri does contain the query string -

> The complete request URI, containing both path and query string. The URI is what was on the status line after the request method. E.g. in "GET /foo/bar?q=s HTTP/1.1" the URI should be /foo/bar?q=s. It could be absolute, some clients send absolute URLs, especially proxies, e.g. http://www.example.org/foo/bar?q=s.

I think if we used `path` we might be good

AndyHWChung: ah ok, let's use "path" then I think to exclude that possibility and be safer

PR #105, opened 2021-12-16
Company matching audit events

AndyHWChung:


app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -584,7 +609,7 @@ object CompanyDetailsController {
 
     Form(
       mapping(
-        "enterCtutr" -> nonEmptyText
+        enterCtutrFormKey -> nonEmptyText
```
priyankadhiman3006, line 612: s change not affect the message conf file and the html file as well?

AndyHWChung: I just factored out the same string value into a `val` so I could reference it in the controller as well when trying to get the submitted value in case of a form error. This is a reference to the `val`

priyankadhiman3006 (approved):

PR #109, opened 2022-01-04
Email journey

priyankadhiman3006:
Hi Please find the PR for  the changes regarding email journey.

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmEmailAddressController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmEmailAddressController.scala
@@ -16,6 +16,7 @@
 
 package uk.gov.hmrc.hecapplicantfrontend.controllers
 
+import cats.implicits.catsSyntaxOptionId
 import cats.instances.future._
 import com.google.inject.{Inject, Singleton}
 import play.api.data.{Form, Mapping}
@@ -23,23 +24,25 @@ import uk.gov.hmrc.emailaddress.{EmailAddress => EmailAddressValidation}
 import play.api.i18n.I18nSupport
 import play.api.mvc._
 import uk.gov.hmrc.hecapplicantfrontend.controllers.ConfirmEmailAddressController.{emailAddressForm, getEmailOptions}
-import uk.gov.hmrc.hecapplicantfrontend.controllers.actions.{AuthAction, RequestWithSessionData, SessionDataAction}
-import uk.gov.hmrc.hecapplicantfrontend.models.{ConfirmUserEmail, EmailAddress, HECSession, UserEmailAnswers}
-import uk.gov.hmrc.hecapplicantfrontend.services.JourneyService
+import uk.gov.hmrc.hecapplicantfrontend.controllers.actions.{AuthAction, AuthenticatedRequest, SessionDataAction}
+import uk.gov.hmrc.hecapplicantfrontend.models.{ConfirmUserEmail, EmailAddress, UserEmailAnswers}
+import uk.gov.hmrc.hecapplicantfrontend.services.{EmailVerificationService, JourneyService}
 import uk.gov.hmrc.hecapplicantfrontend.util.{FormUtils, Logging}
 import uk.gov.hmrc.hecapplicantfrontend.views.html
 import uk.gov.hmrc.http.HeaderCarrier
 import uk.gov.hmrc.play.bootstrap.frontend.controller.FrontendController
-import uk.gov.voa.play.form.ConditionalMappings.{mandatoryIfEqual}
+import uk.gov.hmrc.play.http.HeaderCarrierConverter
+import uk.gov.voa.play.form.ConditionalMappings.mandatoryIfEqual
 
 import java.util.Locale
-import scala.concurrent.{ExecutionContext, Future}
+import scala.concurrent.ExecutionContext
 
 @Singleton
 class ConfirmEmailAddressController @Inject() (
   authAction: AuthAction,
   sessionDataAction: SessionDataAction,
   journeyService: JourneyService,
+  emailVerificationService: EmailVerificationService,
   confirmEmailAddressPage: html.ConfirmEmailAddress,
   mcc: MessagesControllerComponents
 )(implicit ec: ExecutionContext)
@@ -67,15 +70,26 @@ class ConfirmEmailAddressController @Inject() (
     val emailOptions = getEmailOptions(ggEmail)
 
     def handleValidEmail(confirmUserEmail: ConfirmUserEmail) = {
-      val emailAns: Option[UserEmailAnswers] = request.sessionData.fold(_.userEmailAnswers, _.userEmailAnswers)
-      val updatedEmailAnswers                = confirmUserEmail.differentEmail match {
-        case Some(email) => emailAns.map(_.copy(emailAddress = email))
-        case None        => emailAns.map(_.copy(emailAddress = confirmUserEmail.ggEmail))
-      }
-      val updatedSession                     = request.sessionData
-        .fold(_.copy(userEmailAnswers = updatedEmailAnswers), _.copy(userEmailAnswers = updatedEmailAnswers))
 
-      updateAndNextJourneyData(routes.ConfirmEmailAddressController.confirmEmailAddress(), updatedSession)
+      val userSelectedEmail                         = confirmUserEmail.differentEmail.getOrElse(confirmUserEmail.ggEmail)
+      val authReq: AuthenticatedRequest[AnyContent] = request.request
+      val headerCarrier: HeaderCarrier              =
+        HeaderCarrierConverter.fromRequestAndSession(request, request.session)
+
+      val result = for {
+        passcodeResult     <- emailVerificationService.requestPasscode(userSelectedEmail)(headerCarrier, authReq)
+        updatedEmailAnswers = Some(UserEmailAnswers(userSelectedEmail, passcodeResult.some, None))
+        updatedSession      =
+          request.sessionData
+            .fold(_.copy(userEmailAnswers = updatedEmailAnswers), _.copy(userEmailAnswers = updatedEmailAnswers))
+        next               <-
+          journeyService.updateAndNext(routes.ConfirmEmailAddressController.confirmEmailAddress(), updatedSession)
+      } yield next
+
+      result.fold(
+        _.doThrow("Could not update session and proceed"),
+        Redirect
+      )
     }
 
     emailAddressForm(emailOptions)
@@ -94,20 +108,6 @@ class ConfirmEmailAddressController @Inject() (
 
   }
 
-  private def updateAndNextJourneyData(current: Call, updatedSession: HECSession)(implicit
-    r: RequestWithSessionData[_],
-    hc: HeaderCarrier
-  ): Future[Result] =
-    journeyService
-      .updateAndNext(
-        current,
-        updatedSession
-      )
-      .fold(
-        _.doThrow("Could not update session and proceed"),
-        Redirect
-      )
-
 }
 
 object ConfirmEmailAddressController {
@@ -124,7 +124,8 @@ object ConfirmEmailAddressController {
     )
     .verifying(
       Constraint[EmailAddress]((email: EmailAddress) =>
-        if (EmailAddressValidation.isValid(email.value)) Valid
+        if (EmailAddressValidation.isValid(email.value) && email.value.length <= 256) Valid
+        else if (email.value.length > 256) Invalid("error.tooManyChar")
         else Invalid("error.invalidFormat")
       )
     )
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmEmailAddressController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmEmailAddressController.scala
@@ -17,6 +17,7 @@
 package uk.gov.hmrc.hecapplicantfrontend.controllers
 
 import cats.implicits.catsSyntaxOptionId
+import uk.gov.hmrc.hecapplicantfrontend.services.EmailVerificationService
 import cats.instances.future._
 import com.google.inject.{Inject, Singleton}
 import play.api.data.{Form, Mapping}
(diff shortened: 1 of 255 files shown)


app/uk/gov/hmrc/hecapplicantfrontend/services/SendEmailService.scala
```diff
@@ -0,0 +1,70 @@
...
+class SendEmailServiceImpl @Inject() (emailSendConnector: SendEmailConnector, configuration: Configuration)(implicit
+  ec: ExecutionContext
+) extends SendEmailService {
+  override def sendEmail(emailAddress: EmailAddress, emailParameters: EmailParameters)(implicit
+    hc: HeaderCarrier,
+    r: AuthenticatedRequest[_]
+  ): EitherT[Future, Error, EmailSendResult] = {
+    val result: EitherT[Future, Error, HttpResponse] = for {
+      lang      <- EitherT.fromEither[Future](Language.fromRequest(r)).leftMap(Error(_))
+      templateId = if (lang.code === English.code) configuration.get[String]("email-send.template-id-en")
```
AndyHWChung, line 53: this would be safer as a pattern match over the sealed trait in case we ever handle any other languages. We could also move the template id's as `vals` outside of this `def` so we don't keep doing `configuration.get`

app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmEmailAddressController.scala
```diff
@@ -0,0 +1,152 @@
...
+  confirmEmailAddressPage: html.ConfirmEmailAddress,
+  mcc: MessagesControllerComponents
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc)
+    with I18nSupport
+    with Logging {
+
+  val confirmEmailAddress: Action[AnyContent] = authAction.andThen(sessionDataAction).async { implicit request =>
+    request.sessionData.ensureGGEmailIdPresent { ggEmail =>
+      val emailOptions: List[EmailType]                = getEmailOptions
```
AndyHWChung, line 56: doesn't seem worth it to stick a `val` into another `val` when there's no difference between the two. You could just have `getEmailOptions` and maybe rename it to something like `emailTypeOptions`

app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmEmailAddressController.scala
```diff
@@ -0,0 +1,152 @@
...
+    val emailOptions = getEmailOptions
+    request.sessionData.ensureGGEmailIdPresent { ggEmail =>
+      def handleValidEmail(userSelectedEmail: UserSelectedEmail): Future[Result] = {
+        val authReq: AuthenticatedRequest[AnyContent] = request.request
+        val headerCarrier: HeaderCarrier              =
+          HeaderCarrierConverter.fromRequestAndSession(request, request.session)
+
+        val result = for {
+          passcodeResult     <-
+            emailVerificationService.requestPasscode(userSelectedEmail.emailAddress)(headerCarrier, authReq)
```
AndyHWChung, line 81: if you make `requestPasscode` take in `RequestWithSessionData` implicitly instead of `AuthenticatedRequest` then you wouldn't need to pass the implicits explicitly here. When you need the `AuthenticatedRequest` in the service you could just pull it out from the `RequestWithSessionData`

app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmEmailAddressController.scala
```diff
@@ -0,0 +1,152 @@
...
+}
+
+object ConfirmEmailAddressController {
+
+  import play.api.data.validation.{Constraint, Invalid, Valid}
+  import play.api.data.Forms.{mapping, nonEmptyText, of}
+
+  val getEmailOptions: List[EmailType] = List(EmailType.GGEmail, EmailType.DifferentEmail)
+
+  def differentEmailAddressMapping: Mapping[EmailAddress] = nonEmptyText
```
AndyHWChung, line 123: this could be a `val` I think

app/uk/gov/hmrc/hecapplicantfrontend/controllers/VerifyEmailPasscodeController.scala
```diff
@@ -0,0 +1,124 @@
...
+object VerifyEmailPasscodeController {
+
+  def verifyPasscodeForm(): Form[Passcode] = Form(
+    mapping(
+      "passcode" -> nonEmptyText
+        .transform[Passcode](p => Passcode(p.toUpperCase(Locale.UK)), _.value)
+    )(identity)(Some(_))
+  )
+
+  def fetchUserSelectedEmail(session: HECSession): UserSelectedEmail =
```
AndyHWChung, line 113: since this is used elsewhere it could go in a more general place - we could create a method similar to `ensureGGEmailIdPresent` in `HECSession` for example

app/uk/gov/hmrc/hecapplicantfrontend/controllers/EmailAddressConfirmedController.scala
```diff
@@ -0,0 +1,97 @@
...
+  timeProvider: TimeProvider,
+  emailAddressConfirmedPage: html.EmailAddressConfirmed,
+  mcc: MessagesControllerComponents
+)(implicit ec: ExecutionContext)
+    extends FrontendController(mcc)
+    with I18nSupport
+    with Logging {
+
+  val emailAddressConfirmed: Action[AnyContent] = authAction.andThen(sessionDataAction) { implicit request =>
+    val userSelectedEmail = fetchUserSelectedEmail(request.sessionData)
```
AndyHWChung, line 50: I think there's a case for checking that the passcode check result in session is actually positive as well to guard against the case where a user goes to this page directly by entering the url in their browser after they've just selected an email. If it's anything other than the result we want then we could just `sys.error` cos we don't expect it here. The same holds for the submit endpoint below as well

app/uk/gov/hmrc/hecapplicantfrontend/controllers/EmailAddressConfirmedController.scala
```diff
@@ -0,0 +1,97 @@
...
+      val userSelectedEmail                         = fetchUserSelectedEmail(request.sessionData)
+      val authReq: AuthenticatedRequest[AnyContent] = request.request
+      val headerCarrier: HeaderCarrier              =
+        HeaderCarrierConverter.fromRequestAndSession(request, request.session)
+
+      val existingUserEmailAnswers = request.sessionData.userEmailAnswers
+      val emailParameters          = getEmailParameters(request.sessionData)
+
+      val result = for {
+        result             <- sendEmailService.sendEmail(userSelectedEmail.emailAddress, emailParameters)(headerCarrier, authReq)
```
AndyHWChung, line 67: similar to before you could just make `sendEmail` take in a `RequestWithSessionData` implicitly instead of an `AuthenticatedRequest`

app/uk/gov/hmrc/hecapplicantfrontend/controllers/EnterEmailAddressController.scala
```diff
@@ -0,0 +1,120 @@
...
+    with Logging {
+
+  val enterEmailAddress: Action[AnyContent] = authAction.andThen(sessionDataAction).async { implicit request =>
+    val userEmailAnswerOpt: Option[UserEmailAnswers] = request.sessionData.userEmailAnswers
+    val back                                         = journeyService.previous(routes.EnterEmailAddressController.enterEmailAddress())
+    val currentSession                               = request.sessionData
+    val form                                         =
+      userEmailAnswerOpt.fold(enterEmailAddressForm)(uea => enterEmailAddressForm.fill(uea.userSelectedEmail))
+    //Unsetting the user Email answers
+    val updatedSession                               =
```
AndyHWChung, line 59: do the session email answers need to reset here? Couldn't it just happen on the submit?

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -435,6 +477,49 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
       }
     }
 
+  def emailVerificationRoute(session: HECSession): Call = {
+    val emailOpt = session.fold(_.loginData.emailAddress, _.loginData.emailAddress)
+    emailOpt match {
+      case Some(email) if EmailAddress.isValid(email.value) =>
```
AndyHWChung, line 483: we could make sure we only write an email address into `loginData.emailAddress` by doing a check before we write it into session in the `StartController` - something like `.filter(EmailAddress.isValid)` somewhere on the `Option`. This way we wouldn't need to check it here and we also wouldn't need to worry about whether or not it's valid when checking to see if it's present on some of the controller endpoints - I don't think that extra check is being done at the moment on those controller enpoints.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -118,6 +129,23 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
   override def updateAndNext(current: Call, updatedSession: HECSession)(implicit
     r: RequestWithSessionData[_],
     hc: HeaderCarrier
+  ): EitherT[Future, Error, Call] =
+    //if isEmailRequested, then no need to uplift data and no logic needed to go to CYA page
+    if (updatedSession.isEmailRequested) getNextCallAfterTaxCheckPage(current, updatedSession)
```
AndyHWChung, line 134: we probably don't need the extra method here - it could have just gone through the existing logic and arrived at the same result. Another option could have been to leave the existing methods but just maybe put something extra in `upliftToCompleteAnswersIfComplete` to ignore the existing logic if there are already complete answers in session  and just return the existing session.What do you think?

app/uk/gov/hmrc/hecapplicantfrontend/controllers/EnterEmailAddressController.scala
priyankadhiman3006: i have to change it now cause earlier  we thought the link click on passcode verify page should come here. Thanks for pointing

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
priyankadhiman3006: The validity of the email is being tested only here  as i thought if checked at entry point, no need to check at other places. But you are right we can do at start Controller.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
priyankadhiman3006: I think i tried this and it was not working. Let me have a go at this again(as i forget now what happened :)) and will let you know

conf/messages
```diff
@@ -328,3 +332,46 @@ accessDenied.p2 = If you need any assistance accessing this link, please email:
 accessDenied.p3 = <a href="mailto:zaynah.hawa@digital.hmrc.gov.uk?subject=Private Beta link assistance">zaynah.hawa@digital.hmrc.gov.uk</a>
 accessDenied.p4 = <a href="mailto:alex.porter@digital.hmrc.gov.uk?subject=Private Beta link assistance">alex.porter@digital.hmrc.gov.uk</a>
 
+
+# Confirm email address
+confirmEmailAddress.title = Which email address would you like to use?
+confirmEmailAddress.hint = Your tax check code will be sent to this email, any other email addresses you have saved with us will not be changed
+confirmEmailAddress.ggEmail = {0}
```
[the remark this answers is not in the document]
priyankadhiman3006: I tried to pick the ggEmail directly as well but was not able to test that.
In the unit test case when we check the selected radio button ,the index should be 0 or 1 . It was not picking the index. This page was a little complicated one ,  may be at that time some thing was missing , or if there is a better way to do it , suggestions are welcome :)

app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmEmailAddressController.scala
```diff
@@ -53,32 +51,30 @@ class ConfirmEmailAddressController @Inject() (
...
   }
 
   val confirmEmailAddressSubmit: Action[AnyContent] = authAction.andThen(sessionDataAction).async { implicit request =>
-    val emailOptions = getEmailOptions
     request.sessionData.ensureGGEmailIdPresent { ggEmail =>
       def handleValidEmail(userSelectedEmail: UserSelectedEmail): Future[Result] = {
-        val authReq: AuthenticatedRequest[AnyContent] = request.request
-        val headerCarrier: HeaderCarrier              =
-          HeaderCarrierConverter.fromRequestAndSession(request, request.session)
+        //val authReq: AuthenticatedRequest[AnyContent] = request.request
```
AndyHWChung, line 71: this could be cleaned up

conf/application.conf
```diff
@@ -101,6 +101,11 @@ microservice {
       port = 10109
     }
 
+    email-send {
```
AndyHWChung, line 104: we should raise a PR in app-config-base to make it point to the right thing in the test environments

app/uk/gov/hmrc/hecapplicantfrontend/controllers/ConfirmEmailAddressController.scala
priyankadhiman3006: yup forgot 🤦‍♀️

priyankadhiman3006: [updated](https://github.com/hmrc/hec-applicant-frontend/pull/109/commits/3aa4ccebdab0b3aa7f4121443bb5d57a64c5e9e8)

test/uk/gov/hmrc/hecapplicantfrontend/services/SendEmailServiceImplSpec.scala
```diff
@@ -0,0 +1,176 @@
...
+import uk.gov.hmrc.hecapplicantfrontend.models.emailSend.{EmailParameters, EmailSendRequest, EmailSendResult}
+import uk.gov.hmrc.hecapplicantfrontend.models.emailVerification.Language.{English, Welsh}
+import uk.gov.hmrc.hecapplicantfrontend.models.emailVerification.{Passcode, PasscodeRequestResult, PasscodeVerificationResult}
+import uk.gov.hmrc.hecapplicantfrontend.utils.Fixtures
+import uk.gov.hmrc.http.{HeaderCarrier, HttpResponse}
+
+import scala.concurrent.ExecutionContext.Implicits.global
+import scala.concurrent.Future
+
+class SendEmailServiceImplSpec extends AnyWordSpec with Matchers with MockFactory with ControllerSpec {
```
AndyHWChung, line 43: do you need `ControllerSpec` for this service spec here or do you just need the stuff in `PlaySupport`?

priyankadhiman3006: whoops,  just the play support, will change it

priyankadhiman3006: fixed

AndyHWChung (approved):

priyankadhiman3006:
@AndyHWChung just added two test cases for email address confirmed page, where we have added a check that passcode verification result should be a match, otherwise technical error page
[changes here](https://github.com/hmrc/hec-applicant-frontend/pull/109/commits/09ed2fe82ebd9995a3c634aa6311f184143fdbab)

AndyHWChung:
> @AndyHWChung just added two test cases for email address confirmed page, where we have added a check that passcode verification result should be a match, otherwise technical error page [changes here](https://github.com/hmrc/hec-applicant-frontend/pull/109/commits/09ed2fe82ebd9995a3c634aa6311f184143fdbab)

:+1: looks good

PR #110, opened 2022-01-05
audit exit events

AndyHWChung:


app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -57,22 +58,28 @@ trait JourneyService {
...
   implicit val callEq: Eq[Call] = Eq.instance(_.url === _.url)
 
   // map representing routes from one page to another when users submit answers. The keys are the current page and the
   // values are the destination pages which come after the current page. The destination can sometimes depend
   // on state (e.g. the type of user or the answers users have submitted), hence the value type `HECSession => Call`
-  lazy val paths: Map[Call, HECSession => Call] = Map(
+  private def paths(implicit
+    r: Request[_],
+    hc: HeaderCarrier,
+    enableAuditing: EnableAuditing = EnableAuditing(false)
```
AndyHWChung, line 81: I made it so that the audit events triggered from some of the routes defined here.`paths` is used in multiple places but we only want the audit events to trigger when the forward route is being determined from `updateAndNext` - this `EnableAuditing` thing was the lowest impact thing I could think of to selectively audit - let me know if you can think of a better way

app/uk/gov/hmrc/hecapplicantfrontend/models/AuditEvent.scala
```diff
@@ -99,4 +99,73 @@ object AuditEvent {
...
+      val serviceExitReason: String = "SAUTRNotFound"
+
+      val serviceExitDescription: String = "SA UTR not found for the Applicant's NINO"
+    }
+
+    final case class SANoNoticeToFileOrTaxReturn(taxCheckSessionData: HECSession) extends TaxCheckExit {
+      val serviceExitReason: String = "SANoNoticeToFileOrTaxReturn"
+
+      val serviceExitDescription: String =
+        "For relevant income tax year, Self Assessment Notice to File not found, Self Assessment Tax Return not found"
```
priyankadhiman3006, line 134: Should we say like `Either Self Assessment Notice to File or Self Assessment Tax Return not found.` cause it seems there is an or here `SANoNoticeToFileOrTaxReturnllike`

app/uk/gov/hmrc/hecapplicantfrontend/models/AuditEvent.scala
```diff
@@ -99,4 +99,73 @@ object AuditEvent {
...
+    }
+
+    final case class CTEnteredCTUTRNotMatchingBlocked(taxCheckSessionData: HECSession) extends TaxCheckExit {
+      val serviceExitReason: String = "CTEnteredCTUTRNotMatchingBlocked"
+
+      val serviceExitDescription: String =
+        "Applicant has made repeated attempts to provide a matching CT UTR. Attempts limit reached, " +
+          "so Applicant temporarily blocked from making an Application for that CRN"
```
priyankadhiman3006, line 142: so Applicant `is` temporarily - minor typo

app/uk/gov/hmrc/hecapplicantfrontend/models/AuditEvent.scala
```diff
@@ -99,4 +99,73 @@ object AuditEvent {
...
+    }
+
+    final case class CTNoNoticeToFileOrTaxReturn(taxCheckSessionData: HECSession) extends TaxCheckExit {
+      val serviceExitReason: String = "CTNoNoticeToFileOrTaxReturn"
+
+      val serviceExitDescription: String =
+        "For relevant accounting period, Corporation Tax Notice to File not found, Corporation Tax Return not found"
```
priyankadhiman3006, line 149: same here like SA?

app/uk/gov/hmrc/hecapplicantfrontend/models/AuditEvent.scala
AndyHWChung: it's an `and` here so both no return was found and no notice to file was issued. I copied these over from the table so changes would need to be agreed with David - we could do a second pass later on if we wanted to maybe review the descriptions a bit before finalising it - the main thing there was to get the audit events happening in the right place.

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
priyankadhiman3006: yup this seems like a correct place to enable auditing 👍

app/uk/gov/hmrc/hecapplicantfrontend/services/JourneyService.scala
```diff
@@ -134,14 +141,18 @@ class JourneyServiceImpl @Inject() (sessionStore: SessionStore)(implicit ex: Exe
...
-                           upliftedSession.userAnswers.foldByCompleteness(
-                             _ =>
-                               if (currentPageIsCYA) sys.error("All user answers are not complete")
-                               else paths.get(current).map(_(upliftedSession)),
-                             _ =>
-                               if (currentPageIsCYA || upliftedSession.isEmailRequested) paths.get(current).map(_(upliftedSession))
-                               else Some(routes.CheckYourAnswersController.checkYourAnswers())
-                           ),
+                           {
+                             implicit val enableAuditing: EnableAuditing = EnableAuditing(true)
```
priyankadhiman3006, line 145: just a thought should we read the flag for auditing through config instead of hardcoding it as true here?

app/uk/gov/hmrc/hecapplicantfrontend/models/AuditEvent.scala
AndyHWChung: not sure how important the `is` is here - we can let David finalise the descriptions again I think and make changes once they're concrete

priyankadhiman3006 (approved):

PR #112, opened 2022-01-06
bug fix for passcode verificaion no match

priyankadhiman3006:


app/uk/gov/hmrc/hecapplicantfrontend/controllers/VerifyEmailPasscodeController.scala
```diff
@@ -73,14 +75,44 @@ class VerifyEmailPasscodeController @Inject() (
               .map(_.copy(passcode = passcode.some, passcodeVerificationResult = passcodeVerificationResult.some))
           updatedSession              =
             session.fold(_.copy(userEmailAnswers = updatedEmailAnswers), _.copy(userEmailAnswers = updatedEmailAnswers))
-          next                       <-
-            journeyService.updateAndNext(routes.VerifyEmailPasscodeController.verifyEmailPasscode(), updatedSession)
-
-        } yield next
-
-        result.fold(
+          nextOrNoMatch               = passcodeVerificationResult match {
```
AndyHWChung, line 78: you may be able to do something like this instead to make it a bit neater:
```scala
val result = for {
   ...
  nextOrMatch <- passcodeVerificationResult match {
    case PasscodeVerificationResult.NoMatch => EitherT.pure(Left(PasscodeVerificationResult.NoMatch))
    case _ =>  journeyService.updateAndNext(
                          routes.VerifyEmailPasscodeController.verifyEmailPasscode(),
                          updatedSession
                        ).map(Right(_))
  } yield nextOrMatch

result.fold(
  _.doThrow("Could not update session and proceed"),
 _.fold(
   _ => Ok(
      verifyPasscodePage(
          verifyPasscodeForm().withError("passcode", "error.noMatch"),
          journeyService.previous(routes.VerifyEmailPasscodeController.verifyEmailPasscode()),
          userSelectedEmail.emailAddress,
          isGGEmailInSession
         )
     ),
  Redirect
)
```
Then you wouldn't need the recursion either

app/uk/gov/hmrc/hecapplicantfrontend/controllers/VerifyEmailPasscodeController.scala
```diff
@@ -73,14 +77,39 @@ class VerifyEmailPasscodeController @Inject() (
               .map(_.copy(passcode = passcode.some, passcodeVerificationResult = passcodeVerificationResult.some))
           updatedSession              =
             session.fold(_.copy(userEmailAnswers = updatedEmailAnswers), _.copy(userEmailAnswers = updatedEmailAnswers))
-          next                       <-
-            journeyService.updateAndNext(routes.VerifyEmailPasscodeController.verifyEmailPasscode(), updatedSession)
-
-        } yield next
+          nextOrNoMatch              <- passcodeVerificationResult match {
+                                          case PasscodeVerificationResult.NoMatch =>
+                                            val noMatch
```
priyankadhiman3006, line 82: code was not able to interpret the  return type , so i have to explicitly mention it here and at line no 87

AndyHWChung, line 82: can it not determine the correct type without an explicit type here and below?

AndyHWChung: another way to tackle it could be to bring it out into a different method like:
```scala
def f(verificationResult: PasscodeVerificationResult): EitherT[Future, Error, Either[NoMatch.type, Call]] = ...
```
Then you wouldn't need to explicitly give the types on those lines. Not sure what you'd call `f` though :thinking:

priyankadhiman3006: yup i can try that

priyankadhiman3006: [updated](https://github.com/hmrc/hec-applicant-frontend/pull/112/commits/d97a48fba9341e28c08af43fadb007f1265b5e60)

AndyHWChung (approved):

PR #121, opened 2022-01-12
lib update

priyankadhiman3006:


AndyHWChung (commented):
would the tests break if we upgraded to v5.2.0 of scalamock? Not sure if v5 introduces changes that would break us - if it does then we may be able to upgrade to v4.4.0 which is the latest of v4

AndyHWChung (commented):
we could maybe experiment to see if cats 2.4.0 -> 2.7.0 would break anything as well. It'd be worth going through the dependencies and seeing what the latest versions of them are and if they work

priyankadhiman3006:
> we could maybe experiment to see if cats 2.4.0 -> 2.7.0 would break anything as well. It'd be worth going through the dependencies and seeing what the latest versions of them are and if they work

cats is already on 2.7.0, i updated it last time

AndyHWChung:
> > we could maybe experiment to see if cats 2.4.0 -> 2.7.0 would break anything as well. It'd be worth going through the dependencies and seeing what the latest versions of them are and if they work
> 
> cats is already on 2.7.0, i updated it last time

oh yeah, not sure what I was looking at :thinking:

project/AppDependencies.scala
```diff
@@ -3,9 +3,9 @@ import sbt._
 
 object AppDependencies {
 
-  val bootStrapVersion = "5.17.0"
+  val bootStrapVersion = "5.18.0"
   val monocleVersion   = "2.1.0"
```
AndyHWChung, line 7: is v3.1.0 of monocle possible?

priyankadhiman3006: the latest is 
3.0.0-M6, all version after 2.1.0 are like this , ii thought may be they are nit that stable to stuck to 2.1.0, i can double check

priyankadhiman3006: plus after 2.1.0, all versions are compatible with 2.13 scala only and not with 2.12

AndyHWChung: ah ok :+1:

AndyHWChung (approved):

PR #123, opened 2022-01-13
look back period logic change

priyankadhiman3006:
Updated the look back period logic with test cases. 
-Some refactoring is also done for the test cases, Rest of the refactoring will do in separate PR

### DIFF
--- a/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
+++ b/app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
@@ -590,7 +590,7 @@ object CompanyDetailsController {
     */
   def calculateLookBackPeriod(today: LocalDate): (LocalDate, LocalDate) = {
     val currentDay = if (today.getMonth.getValue === 2 && today.getDayOfMonth === 29) today.plusDays(1) else today
-    currentDay.minusYears(2).plusDays(1) -> currentDay.minusYears(1)
+    currentDay.minusYears(2) -> currentDay.minusYears(1).minusDays(1)
   }
 
   val enterCtutrFormKey: String = "enterCtutr"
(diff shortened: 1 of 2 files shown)


app/uk/gov/hmrc/hecapplicantfrontend/controllers/CompanyDetailsController.scala
```diff
@@ -590,7 +590,7 @@ object CompanyDetailsController {
     */
   def calculateLookBackPeriod(today: LocalDate): (LocalDate, LocalDate) = {
     val currentDay = if (today.getMonth.getValue === 2 && today.getDayOfMonth === 29) today.plusDays(1) else today
-    currentDay.minusYears(2).plusDays(1) -> currentDay.minusYears(1)
+    currentDay.minusYears(2) -> currentDay.minusYears(1).minusDays(1)
```
AndyHWChung, line 593: is the logic for `currentDay` above still necessary? The rule only cares about if the day before is on the 28th or 29th of Feb rather than today

priyankadhiman3006: i think yes , if today is 29, it should treat today as 1 march and then calculate the rest. otherwise it will give the dates as (2022-02-28,2023-02-27) for 29 feb 2024

priyankadhiman3006: whoops i think you are right, it should behave as scenario1 , we will double confirm later, till then i can remove this logic

AndyHWChung: ah ok, it'd be worth double checking to see if that's the correct rule now still just in case. It's not an example on that spreadsheet so it's not 100% clear what the correct lookup period should be.

priyankadhiman3006: yeah, i guess we can keep the PR open till then

AndyHWChung (approved):

PR #150, opened 2022-02-02
HEC-1780 exit page for bad emil request

priyankadhiman3006:
ticket details -> https://jira.tools.tax.service.gov.uk/browse/HEC-1780

### DIFF
--- a/test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
+++ b/test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
@@ -3446,6 +3446,47 @@ class JourneyServiceImplSpec extends ControllerSpec with SessionSupport with Aud
 
         }
 
+        "the cannot send verification passcode email page" when {
+
+          def test(emailAddress: Option[EmailAddress], previousRoute: Call) = {
+            val session = Fixtures.individualHECSession(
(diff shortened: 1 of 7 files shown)


test/uk/gov/hmrc/hecapplicantfrontend/services/JourneyServiceImplSpec.scala
```diff
@@ -3446,6 +3446,47 @@ class JourneyServiceImplSpec extends ControllerSpec with SessionSupport with Aud
 
         }
 
+        "the cannot send verification passcode email page" when {
```
AndyHWChung, line 3449: I think it'd be worth having unit tests to check that the call to `.previous` works when the current page is `routes.ConfirmEmailAddressController.confirmEmailAddress()` or `routes.EnterEmailAddressController.enterEmailAddress()` and the session has the bad email request state in it

priyankadhiman3006: The code logic is when current page is either routes.ConfirmEmailAddressController.confirmEmailAddress() or routes.EnterEmailAddressController.enterEmailAddress(), back always goes to tax check complete page irrespective of the other values to avoid loop. It was discussed ages ago not with respect to this page but some other page. If that logic needs to be changed please let me know

AndyHWChung (approved):

PR #157, opened 2022-02-15
HEC-1849 exit survey links on more pages

nubz:


app/uk/gov/hmrc/hecapplicantfrontend/views/CannotDoTaxCheck.scala.html
```diff
@@ -28,4 +29,7 @@
 @layout(pageTitle = Some(title), backLocation = Some(back)) {
     <h1 class="govuk-heading-xl">@title</h1>
     <p class="govuk-body govuk-!-margin-bottom-9">@Html(messages(s"$key.p1", appConfig.contactHmrcSa))</p>
+    <p class="govuk-body">
```
AndyHWChung, line 32: is it worth having a small separate component which contains this so we can just put `@exitSurvey()` here and in other places? If we ever have to change the way it looks or update the url then we'd just have to update in one place if we do that

nubz: done

PR #166, opened 2022-03-03
HEC-1916 pull licence type and expiry date into email sent page

nubz:


app/uk/gov/hmrc/hecapplicantfrontend/controllers/EmailSentController.scala
```diff
@@ -35,7 +37,19 @@ class EmailSentController @Inject() (
 
   val emailSent: Action[AnyContent] = authAction.andThen(sessionDataAction) { implicit request =>
     request.sessionData.ensureUserSelectedEmailPresent { userSelectedEmail =>
-      Ok(emailSentPage(userSelectedEmail.emailAddress))
+      val licenceType = request.sessionData.userAnswers.foldByCompleteness(
```
AndyHWChung, line 40: this may not be the best way to get the licence type given that we are planning to send emails off the display tax check codes page - from that page we won't have complete answers in session. There's another way we can make this information available (as well as the tax check code) but it needs some changes to the session data structure - I'm happy to do that change now so that we're ready for the future

nubz: Yes good thinking, obvs I used what was available in session but if you are happy to build a session data structure for existing tax check codes we can use that

AndyHWChung: just updated it [here](https://github.com/hmrc/hec-applicant-frontend/pull/166/commits/b011a2eb6dd39464fdb9101adb968204b8637a61) - I'll see if I can get Priyanka to review

AndyHWChung (approved):

priyankadhiman3006 (approved):

PR #188, opened 2022-04-01
HEC-2026 custom UR banner

nubz:


app/uk/gov/hmrc/hecapplicantfrontend/views/components/CustomHmrcStandardHeader.scala.html
```diff
@@ -0,0 +1,51 @@
...
+        serviceName: Option[String] = None,
+        serviceUrl: Option[String] = None,
+        signOutUrl: Option[String] = None,
+        userResearchBanner: Option[UserResearchBanner] = None,
+        phaseBanner: Option[PhaseBanner] = None,
+        displayHmrcBanner: Boolean = false,
+        additionalBannersBlock: Option[Html] = None,
+        containerClasses: String = Header.defaultObject.containerClasses
+)(implicit messages: Messages)
+@defining("service.name") { serviceNameKey =>
```
AndyHWChung, line 35: `@defining` is a bit hard to look at - does this not work by assigning vals instead this?
```scala
@serviceNameKey = @{ "service.name" }
@serviceNameFromMessages = @{  if(messages.isDefinedAt(...) }
```

nubz: yes - I will update, we are deviating more and more from library implementation though, but hopefully library updates will only need to be patchd in for as long as this survey is alive and then we can revert back to the library components

AndyHWChung (approved):

PR #200, opened 2022-05-13
HEC-2121 stop using fat arrow functions

nubz:


AndyHWChung (approved):
just need to make sure the commit message has 2122 as the ticket number rather than 2121

PR #203, opened 2022-05-26
HEC-2125: fix back link on entity type exit page

AndyHWChung:


app/uk/gov/hmrc/hecapplicantfrontend/controllers/actions/UncertainEntityTypeJourneyAction.scala
```diff
@@ -80,3 +87,23 @@ class UncertainEntityTypeJourneyAction @Inject() (
       .merge
   }
 }
+
+class RequireDidConfirmUncertainEntityTypeJourneyAction @Inject() (
```
priyankadhiman3006, line 91: can we combine this class and the class below to add one param requireDidConfirmUncertainEntityType? As both class looks the same except for this val

priyankadhiman3006 (approved):

app/uk/gov/hmrc/hecapplicantfrontend/controllers/actions/UncertainEntityTypeJourneyAction.scala
AndyHWChung: I tried to do something like this but the class gets injected into the controllers as an input parameter and I couldn't find a way to specify the value of the flag at those points. When the injected value gets used there doesn't seem to be a way to pass in the value of the flag when you use it cos the value itself is used

PR #217, opened 2022-06-22
HEC-2199: don't filter engwal data

AndyHWChung:


conf/application.conf
```diff
@@ -225,4 +225,12 @@ email-send {
 
 accessibility-statement.service-path = "/tax-check-for-licence"
 
-scot-ni-email-allow-list = [ "user@test.com" ]
+scot-ni-email-allow-list = [
+  "user@test.com",
+  "user1@test.com"
+]
+
+scot-ni-email-allow-list-is-england-or-wales = [
```
AndyHWChung, line 233: it's a bit ugly having separate lists in config like this instead of something like
```
scot-ni-email-allow-list = [
  {
    email = "..."
    isEnglandOrWales = true
  }, 
  {
    email = "..."
    isEnglandOrWales = false
  },
]
```
but I'm not convinced that the syntax in the app-config files allows for a list of objects

priyankadhiman3006: Can we have a  something like ""user@test.com, false", and then we can split the string by del ","?

priyankadhiman3006: ok, then it's fine 👍

priyankadhiman3006 (approved):

